PR #613, opened 2015-05-13
Write/WriteHeader are modifiable more than once

EmperorEarth:
Modified logRecord's Write & WriteHeader methods to mimic net/http's …functionality, e.g. modifiable only once

First time contributing to someone else's project and a new gopher. Take your time to make sure it's a good change!


bradfitz:
We don't use Github pull requests for code review. See https://camlistore.org/code

We only use Github as a mirror and for its issue tracker.

/cc @mpl

PR #633, opened 2015-07-29
Fix small grammar error

rychipman:
Came across this while reading the camlistore overview. I think it's a good habit to actually point out little errors when I come across them in public-facing documents, so even though it's a one-line patch, here it is!


mpl:
Please read http://camlistore.org/code#contributing

PR #673, opened 2016-01-14
pass mysql host to camtool dbinit

rfistman:
dbinit for mysql wasn't using the `host`. This works for my case, but I don't know if I've broken something.


mpl:
@rfistman We don't do pull requests, please see https://camlistore.org/code#contributing

bradfitz:
Alternatively, sign the CLA at least and ask one of us to submit it for you.

But generally we also require test updates for changes.

Issue #897, opened 2016-12-30
pinentry runPass not disabling echo


gcoxmoz:
I was running a package within docker, and my passphrase kept being echo'ed back at me.  Likewise on OSX.

https://github.com/camlistore/camlistore/blob/master/pkg/misc/pinentry/pinentry.go#L130 obviously calls for disabling echo.

However, in runPass, there's a few simplistic issues.  It was hard to track because:
1a. Nothing looks for any errors from cmd.Run; it just assumes it succeeds.
1b. cmd.Stderr wasn't set equal to os.Stderr.

After tweaking those, I found runPass was getting "stty: stdin isn't a terminal".
2. With a simple add of cmd.Stdin = os.Stdin it went to behaving, and my passphrase wasn't echo'ed.

I'm not offering this as a PR specifically because it feels like it's one of those "trivial to reconstruct" things if I'm on the right track; but that it's been around for multiple years makes me wonder if I'm missing something obvious here that would make my naive PR the wrong approach.

mpl:
@gcoxmoz thanks.

@bradfitz wrote that part, and I've never used it, so I'm not sure about the original intent. I'll have a better look at it later if Brad does not answer.

mpl:
@gcoxmoz sorry about the delay.

I've had a look and I could reproduce, both the problems you describe and test the fixes as well. imho it sgtm, namely, to:

1) make runPass return an error (and check it in the caller), that includes stdout and stderr in the case where cmd.Run didn't return nil. (so you probably want to set both cmd.Stdout and cmd.Sterr to some bytes.Buffer).
2) set cmd.Stdin

Do you want to submit a CL for it?

gcoxmoz:
https://github.com/mozilla/mig/issues/317 has a decently viable patch.  Someone who's set up to jump through the CL hoops can run with it.

mpl:
https://camlistore-review.googlesource.com/9291

PR #906, opened 2017-02-02
pinentry runPass should disable echo

gcoxmoz:
Not really a go person, but, I'm hoping this fixes Issue #897.


mpl:
Hi,

We do not take github PRs, as we work with gerrit.
Please see https://camlistore.org/code#contributing

PR #918, opened 2017-04-16
make: fixed copy/paste by replacing *website with *camnetdns in targets build

nxtvibe:


mpl:
@nxtvibe hey, thanks for the PR, but as gerrithub.io (what we use to import github PRs as gerrit CLs) is pooping itself once more, I can't import your patch at the moment.
If you want to speed things up, you can send a proper CL by following https://camlistore.org/code#contributing . Otherwise, I'll keep on retrying to import it later.

nxtvibe:
Done :)

mpl:
thanks!

PR #950, opened 2017-09-15
travis: update go version

dvrkps:


mpl:
We haven't even tested and migrated Camlistore to 1.9 (https://github.com/camlistore/camlistore/issues/948), so I don't think this is appropriate, is it?

what does 1.x do, what version does it correspond to?

dvrkps:
@mpl 1.x match latest stable, in this moment it is 1.9.0

Maybe 1.8.x is better for now ?

mpl:
yes, I suppose that works, if you test - because I havent - first that everything indeed works with 1.8.3 (which should not be a problem).

Also, we don't use github for contributions (https://camlistore.org/code), so could you please send a gerrit CL instead?

PR #1080, opened 2018-03-28
Introduce listenLetsEncrypt option

zoranzaric:
The default for the Let's Encrypt listen address is still ':80'.  This
`camlistored`flag and config option provides a method to bind to a specific IP
or a different port.

For Iss1078


mpl:
We do not use gihub PRs. please see https://perkeep.org/code#contributing

PR #1260, opened 2019-06-06
server/perkeepd/ui: Implement blob audio

omarchehab98:
Adds audio support to the web UI

### DIFF
--- a/server/perkeepd/ui/audio_detail.js
+++ b/server/perkeepd/ui/audio_detail.js
@@ -1,0 +1,98 @@
+/*
+Copyright 2019 The Perkeep Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+goog.provide('cam.AudioDetail');
+
+goog.require('cam.BlobItemAudioContent');
+
+cam.AudioDetail = React.createClass({
+	displayName: 'AudioDetail',
+
+	propTypes: {
+		permanodeMeta: React.PropTypes.object,
+		resolvedMeta: React.PropTypes.object.isRequired,
+	},
+
+	render: function() {
+		return React.DOM.div({
+				className: 'cam-detail-audio',
+			},
+			this.getControls_(),
+		);
+	},
+
+	getControls_: function() {
+		var mediaTags = this.props.resolvedMeta.mediaTags;
+		return React.DOM.div({
+				className: 'cam-detail-audio-controls',
+			},
+			React.DOM.div({
+					className: 'cam-detail-audio-controls-meta'
+				},
+				mediaTags.title != null && React.DOM.div({
+					className: 'cam-detail-audio-controls-meta-title',
+				}, mediaTags.title),
+
+				mediaTags.artist != null && React.DOM.div({
+					className: 'cam-detail-audio-controls-meta-artist',
+				}, mediaTags.artist),
+			),
+
+			React.DOM.div({
+					className: 'cam-detail-audio-controls-player',
+				},
+				React.DOM.audio({
+					controls: true,
+					key: this.props.resolvedMeta.blobRef,
+					src: goog.string.subs('%s%s/%s', goog.global.CAMLISTORE_CONFIG.downloadHelper, this.props.resolvedMeta.blobRef, this.props.resolvedMeta.file.fileName),
+				}),
+			),
+		);
+	},
+});
+
+cam.AudioDetail.getAspect = function(blobref, searchSession) {
+	if (!blobref) {
+		return null;
+	}
+
+	var rm = searchSession.getResolvedMeta(blobref);
+	var pm = searchSession.getMeta(blobref);
+
+	if (!pm) {
+		return null;
+	}
+
+	if (pm.camliType != 'permanode') {
+		pm = null;
+	}
+
+	if (rm && cam.BlobItemAudioContent.isAudio(rm)) {
+		return {
+			fragment: 'audio',
+			title: 'Audio',
+			createContent: function(size, backwardPiggy) {
+				return React.createElement(cam.AudioDetail, {
+					key: 'audio',
+					permanodeMeta: pm,
+					resolvedMeta: rm,
+				});
+			},
+		};
+	} else {
+		return null;
+	}
+};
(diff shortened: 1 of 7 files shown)


server/perkeepd/ui/blob_item_audio_content.js
```diff
@@ -0,0 +1,178 @@
...
+		blobref: React.PropTypes.string.isRequired,
+		filename: React.PropTypes.string.isRequired,
+		href: React.PropTypes.string.isRequired,
+		size: React.PropTypes.instanceOf(goog.math.Size).isRequired,
+	},
+
+	getInitialState: function() {
+		return {
+			mouseover: false,
+			playing: false,
```
omarchehab98: @mpl in response to your comments on Gerrit:

>I think the point was not about the media being loaded, but about the react element itself. imagine there are 50 videos in your current page, then it probably takes a little bit of time for all the buttons to load on the page.
>[Source](https://perkeep-review.googlesource.com/c/perkeep/+/19326/1/server/perkeepd/ui/blob_item_audio_content.js#37)

Since we're using React, buttons are created and added to the DOM on the client-side. It's a safe assumption that your buttons will be rendered soon after your React component is instantiated.

Correction, I said `loaded` is not used in video preview, it is, here is how `loaded` is used in `blob_item_video_content`:
* Adds the class `cam-blobitem-video-loaded` if `loaded`, however, class is not implemented
* Renders `getPoster_` if not `loaded`
* Renders `getVideo_` if `loaded`
* Sets `loaded` to `true` when clicking play or pause button

Basically, loaded is whether or not the video was ever played by the user. We use that information to decide whether to display the poster or the video. Audio has no visual change when it is being played, so we don't need `loaded` here.

server/perkeepd/ui/blob_item_audio_content.js
```diff
@@ -0,0 +1,178 @@
...
+			mouseover: false,
+		});
+	},
+});
+
+cam.BlobItemAudioContent.isAudio = function(rm) {
+	// From https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats
+	var extensions = [
+		'webm',
+		'ogg',
```
omarchehab98: >>Good catch, I didn't know `ogg` is also a video format. Let's ensure video takes precedence
>
>Because you added the audio handler after the video one in BLOB_ITEM_HANDLERS_, I think it's already the case.
>
>>and use metadata to determine whether the blob is audio or video.
>
>yes, that would be nice.
>But on first approximation if you want, we can assume most oggs are in fact audio (which I believe is practically the case), and the video handler can always let the audio handler deal with oggs for now.

TODO: Give audio priority over video for `ogg`

server/perkeepd/ui/blob_item_audio_content.js
mpl: understood.

server/perkeepd/ui/blob_item_audio_content.js
```diff
@@ -0,0 +1,178 @@
...
+		);
+	},
+
+	setAudioRef_: function(audio) {
+		this.audioRef_ = audio;
+	},
+
+	handlePlayPauseClick_: function(e) {
+		if (this.state.playing) {
+			this.audioRef_.pause();
```
mpl: why is it better to keep an audioRef_ around, rather than use ReactDOM.findDOMNode(this).querySelector('audio') to find it, like it was done for video?

server/perkeepd/ui/blob_item_audio.css
```diff
@@ -0,0 +1,46 @@
...
+	text-align: center;
+}
+
+.cam-blobitem-audio-play,
+.cam-blobitem-audio-pause {
+	color: rgba(125, 125, 125, 0.7);
+	position: absolute;
+	left: 50%;
+	top: 50%;
+	transform: translate(-50%, -50%);
```
mpl: what does that do?

server/perkeepd/ui/index.js
```diff
@@ -80,6 +81,7 @@ cam.IndexPage = React.createClass({
 		cam.BlobItemInstapaperContent.getHandler,
 		cam.BlobItemImageContent.getHandler,
 		cam.BlobItemVideoContent.getHandler,
+		cam.BlobItemAudioContent.getHandler,
```
mpl: Just move the audio handler above the video handler, and I think that should be enough to give the audio handler priority for oggs.

omarchehab98:
findDOMNode is deprecated, using it discouraged

server/perkeepd/ui/audio_detail.js
```diff
@@ -0,0 +1,105 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+goog.provide('cam.AudioDetail');
+
+goog.require('cam.BlobItemAudioContent');
+
+// Renders the guts of the detail view for images.
```
mpl: comment needs fixing

server/perkeepd/ui/audio_detail.js
```diff
@@ -0,0 +1,105 @@
...
+cam.AudioDetail = React.createClass({
+	displayName: 'AudioDetail',
+
+	propTypes: {
+		backwardPiggy: React.PropTypes.bool.isRequired,
```
mpl: we don't need the piggy for this aspect, do we?

server/perkeepd/ui/audio_detail.js
```diff
@@ -0,0 +1,105 @@
...
+		height: React.PropTypes.number.isRequired,
```
mpl: height does not seem to be used.

server/perkeepd/ui/audio_detail.js
```diff
@@ -0,0 +1,105 @@
...
+		permanodeMeta: React.PropTypes.object,
+		resolvedMeta: React.PropTypes.object.isRequired,
+		width: React.PropTypes.number.isRequired,
```
mpl: same as for weight

server/perkeepd/ui/audio_detail.js
```diff
@@ -0,0 +1,105 @@
...
+				},
+				React.DOM.audio({
+					controls: true,
+					key: this.props.resolvedMeta.blobRef,
+					src: goog.string.subs('%s%s/%s', goog.global.CAMLISTORE_CONFIG.downloadHelper, this.props.resolvedMeta.blobRef, this.props.resolvedMeta.file.fileName),
+				}),
+			),
+		);
+	},
+});
```
mpl: this looks nice, thanks.

server/perkeepd/ui/blob_item_audio.css
omarchehab98: `left: 50%` and `top: 50%` move the element's top left corner to the center `transform: translate(-50%, -50%)` moves the element 50% of it's width and height towards to the negative `x` and `y` direction.

In english, it centers the element using the element's center as an anchor

mpl: ACK, thx.

omarchehab98:
@mpl sounds good 🚀 

Regarding force push, if you click on the link "force-pushed" it will show a diff of what got amended
Alternatively, if you look at the files changed tab you will be able to see all the changes with the new amends

PR #1263, opened 2019-06-12
server/perkeepd/ui: Indent using tabs

omarchehab98:
Convert all files to tabs instead of spaces to adhere with style guide

### DIFF
--- a/server/perkeepd/ui/mobile.html
+++ b/server/perkeepd/ui/mobile.html
@@ -42,7 +42,7 @@ limitations under the License.
 		<script>
 			var render = function() {
 				var newURL = new goog.Uri(window.location.href);
-			        var MobileSetupView = React.createFactory(cam.MobileSetupView);
+					var MobileSetupView = React.createFactory(cam.MobileSetupView);
 				var mobile = MobileSetupView({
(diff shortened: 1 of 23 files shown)


bradfitz:
Let's not modify svg files.

Which style guide, btw? Link it in commit message?

Also, you can delete the Change-Id. That's no longer needed. What added that? Did you do it by hand? Or you're manually moving over a change from Gerrit?

omarchehab98:
Sure, I'll revert the SVG files.

Whoops, should've mentioned that it is the styleguide found in the doc
folder doc/web-ui-styleguide.md

Change-Id was added automatically by a commit hook, I should remove the
hook since it's not longer needed.

bradfitz:
@googlebot, PTAL.

omarchehab98:
@bradfitz ping

server/perkeepd/ui/mobile.html
```diff
@@ -42,7 +42,7 @@
 		<script>
 			var render = function() {
 				var newURL = new goog.Uri(window.location.href);
-			        var MobileSetupView = React.createFactory(cam.MobileSetupView);
+					var MobileSetupView = React.createFactory(cam.MobileSetupView);
```
mpl: that looks wrong

Issue #54, opened 2012-10-08
drag-and-drop directories in the web UI


bradfitz:
<pre>Chrome now (or will shortly) support drag &amp; drop directories:

<a href="http://code.google.com/p/chromium/issues/detail?id=99823">http://code.google.com/p/chromium/issues/detail?id=99823</a>
<a href="http://wiki.whatwg.org/wiki/DragAndDropEntries">http://wiki.whatwg.org/wiki/DragAndDropEntries</a>

We should add support for this in the Camlistore web UI.</pre>

Issue #657, opened 2015-12-27
UI: drag-and-drop folders


bradfitz:
Use the non-standard Javascript File System API to drag-and-drop folders:

https://developers.google.com/web/updates/2012/07/Drag-and-drop-a-folder-onto-Chrome-now-available?hl=en

https://developer.mozilla.org/en-US/docs/Web/API/File_System_API

http://www.html5rocks.com/en/tutorials/file/filesystem/

Only Chrome, though: http://caniuse.com/#feat=filesystem

mpl:
cc @dotMR

omarchehab98:
Duplicate of #54

PR #1264, opened 2019-06-13
server/perkeepd/ui: Drag-and-drop upload files and directories

omarchehab98:
# Summary:

Re-implement drag-and-drop handler to upload files and optionally support directories if browser implements `webkitGetAsEntry` on `DataTransferItem`.

# Test Plan:

Case A) Directory

1. Drag and drop a directory
2. Perkeep should receive all files, even deeply nested ones

Case B) Files

1. Drag and drop a selection of many files
2. Perkeep should receive all files

Case C) Other elements

1. Drag and drop non-file and non-directory elements such as an HTML image
2. Nothing should happen

# Closes: #54 #657


omarchehab98:
In the second commit, I added abort handlers to server connection methods so that we can implement cancel functionality in the upload dialog. I should have made this it's own pull request tbh, I'll rebase it to onto master and remove it from this PR

omarchehab98:
@bradfitz @mpl could someone please review this?

mpl:
I've just had a quick try, and I see that this PR "flattens" all the directories hierarchy, i.e. it uploads each file as the camliContent of a stand-alone permanode. it's good, but I don't remember what we said last time we discussed it (if we ever), so I'm not sure this is what we want as the default behaviour.

I can see at least 2 other alternatives:
1) we'd still have a permanode per file, but we'd keep the hierarchy through sets, i.e. each directory would be represented by a set (a permanode with a camliMember attribute).
2) we do not create a permanode per file. we faithfully create files and directories schemas, just as things were on the host.

So I think we need to discuss that on the referenced issue before moving forward with that PR.

PR #1265, opened 2019-06-13
pkg/index: wait for the database to be actually ready when creating kv

mpl:
pkg/test/dockertest: publish container port on random host port, because
container IP is actually only reachable when natively on linux. And it
has to be a random, variable, port, otherwise concurrent tests runs would
block each other when trying to bind on same port.

=> makes the MySQL tests in pkg/index pass on darwin.

### DIFF
--- a/pkg/index/mysql_test.go
+++ b/pkg/index/mysql_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package index_test
 
 import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


mpl:
FWIW, I've run "go test ./..." also on my linux machine and tests were passing. So I have to investigate what exactly goes wrong on Travis.

PR #1269, opened 2019-06-24
docs: remove references to SPDY

zenhack:
Which has been deprecated for some time now; the references to it really
date the documents.

### DIFF
--- a/doc/protocol/blob-stat.md
+++ b/doc/protocol/blob-stat.md
@@ -2,7 +2,7 @@
 
 This document describes the "batch stat" API end-point, for checking
 the size/existence of multiple blobs when the client and/or server do
-not support SPDY or HTTP/2.0.  See [blob-upload](blob-upload.md) for more
+not support HTTP 2.0.  See [blob-upload](blob-upload.md) for more
 background.
 
 Notably: the HTTP method may be GET or POST.  GET is more correct but
--- a/doc/protocol/blob-stat.md
+++ b/doc/protocol/blob-stat.md
@@ -2,7 +2,7 @@
 
 This document describes the "batch stat" API end-point, for checking
 the size/existence of multiple blobs when the client and/or server do
-not support HTTP 2.0.  See [blob-upload](blob-upload.md) for more
+not support HTTP/2.0.  See [blob-upload](blob-upload.md) for more
 background.
 
 Notably: the HTTP method may be GET or POST.  GET is more correct but
--- a/doc/protocol/blob-upload.md
+++ b/doc/protocol/blob-upload.md
@@ -13,17 +13,17 @@ Uploading a single blob is done in two parts:
 
 
 When uploading multiple blobs (the common case), the fastest option
-depends on whether or not you're using a modern HTTP transport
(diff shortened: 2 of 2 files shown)


doc/protocol/blob-upload.md
```diff
@@ -13,17 +13,17 @@ Uploading a single blob is done in two parts:
 
 
 When uploading multiple blobs (the common case), the fastest option
-depends on whether or not you're using a modern HTTP transport
-(e.g. SPDY).  If your client and server don't support SPDY, you want
-to use the batch stat and batch upload endpoints, which hopefully can
-die when the future finishes arriving.
+depends on whether or not you're using HTTP 2.0.  If your client and
```
mpl:
```suggestion
depends on whether or not you're using HTTP/2.0.  If your client and
```

mpl: why did you mark it resolved?

zenhack:
Sorry for the slow reply; I was out in the woods all last week.

IIRC, there was a button to auto-apply you suggested changes, and when I
clicked it it also marked it as resolved. The CLA bot seems to have
freaked out about the commit it generated (which had both of us as
authors). It's a feature I've not seen before, perhaps it's new.

I then noticed a couple other places where the same change needed to be
made, and so did that locally and squashed, hence the force push.

Sorry for the confusion.

mpl:
> Sorry for the slow reply; I was out in the woods all last week.

no problem, that's what woods are for.

> IIRC, there was a button to auto-apply you suggested changes, and when I clicked it it also marked it as resolved. The CLA bot seems to have freaked out about the commit it generated (which had both of us as authors). It's a feature I've not seen before, perhaps it's new. I then noticed a couple other places where the same change needed to be made, and so did that locally and squashed, hence the force push. Sorry for the confusion.

Well, I don't know what went wrong in the whole process, but the reason I'm asking is that the current state of the PR is the opposite of what I was suggesting: we need "HTTP/2.0" everywhere, not "HTTP 2.0".

zenhack:
Quoting mpl (2019-07-09 04:05:26)

>    Well, I don't know what went wrong in the whole process, but the
>    reason I'm asking is that the current state of the PR is the
>    opposite of what I was suggesting: we need "HTTP/2.0" everywhere,
>    not "HTTP 2.0".

Aha. Well, my original thought process is swapped out to disk somewhere,
but I pushed a fix.

-Ian

doc/protocol/blob-stat.md
```diff
@@ -2,7 +2,7 @@
 
 This document describes the "batch stat" API end-point, for checking
 the size/existence of multiple blobs when the client and/or server do
-not support SPDY or HTTP/2.0.  See [blob-upload](blob-upload.md) for more
+not support HTTP 2.0.  See [blob-upload](blob-upload.md) for more
```
mpl: almost there :-)
```suggestion
not support HTTP/2.0.  See [blob-upload](blob-upload.md) for more
```

mpl:
I don't understand why the cla check is failing, but I'm going to assume we're all good since you already submitted CLs/PRs to the project in the past.

PR #1274, opened 2019-07-17
String constraints: support regexps.

zenhack:


tgulacsi:
Can you add some tests? Esp. the edge cases (wrong regexp, both regexp and non-regexp specified...)

PR #1277, opened 2019-08-19
WIP: Initial PoC for bulk attributes setting

tomberek:
Needed to do bulk updates of attributes and found the `pk put attr` tool to be slow on multiple invocations. This just streams in argument lines a-la xargs in a for loop and is much faster as a result. This is just meant to be a PoC.


tomberek:
@googlebot I signed it!

Issue #1226, opened 2018-10-01
Support use as a pure metadata/indexing frontend without blob storage


khimaros:
It would be great if users could take advantage of the benefits of Perkeep (indexing, search, out of band metadata, duplicate detection, and remote blob sync) without replicating their entire file-based storage (and incurring the disk-space penalty). A tool which does this very well (for music) is https://github.com/beetbox/beets -- a challenge I foresee with this is that Perkeep blobs are designed to be referenced by checksum.

PR #1282, opened 2019-11-27
pkg/blobserver/fsbacked: add blobserver using existing local files as storage [WIP]

bobg:
This PR is a **preliminary sketch** for a new blobserver type that uses files uploaded to it as their own storage.

When you add a file to an `fsbacked.Storage` that's within the directory tree it controls, an entry is added to a database that maps between files and blobrefs; but the file's contents are not copied anywhere. When fetching the file's content blob later, the database directs the `Storage` to the right local file and the data is served from there.

Adding files _outside_ the directory tree, or adding any other kind of blob, fails over to another blobserver nested inside the `fsbacked.Storage`.

This solves the problem of wanting to add a tree of large files (e.g., videos of my kids growing up) to a local Perkeep instance without storing all the data twice. This should be used only on directory trees whose files do not change, lest the blobrefs in the database become mismatched to their corresponding files.

A number of other changes throughout Perkeep would be needed to make this truly useful. The `io.Reader` presented to a blobserver's `ReceiveBlob` method is usually (always?) some wrapper object (like [checkHashReader](https://github.com/perkeep/perkeep/blob/d342b0e26632217a93a7b9a2ce85acca0c5cd00b/pkg/blobserver/receive.go#L71-L79)) that conceals the underlying `*os.File`, without which `fsbacked.Storage` cannot detect that a file within its tree is being uploaded. And in any case, Perkeep imposes [rather a low limit on blob sizes](https://github.com/perkeep/perkeep/blob/d342b0e26632217a93a7b9a2ce85acca0c5cd00b/pkg/constants/constants.go#L22-L23) for this purpose.

Presented for further discussion.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/lib/pq v1.1.1
 	github.com/mailgun/mailgun-go v0.0.0-20171127222028-17e8bd11e87c
 	github.com/mattn/go-mastodon v0.0.5-0.20190517015615-8f6192e26b66
-	github.com/mattn/go-sqlite3 v1.6.0
+	github.com/mattn/go-sqlite3 v1.10.0
 	github.com/miekg/dns v0.0.0-20161003181808-3f1f7c8ec9ea
 	github.com/nf/cr2 v0.0.0-20140528043846-05d46fef4f2f
 	github.com/onsi/ginkgo v1.8.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/nf/cr2 v0.0.0-20140528043846-05d46fef4f2f
 	github.com/onsi/ginkgo v1.8.0 // indirect
 	github.com/onsi/gomega v1.5.0 // indirect
-	github.com/pkg/errors v0.8.0 // indirect
+	github.com/pkg/errors v0.8.0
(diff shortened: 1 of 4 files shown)


zenhack:
Re: the blob size limit, you could solve that by keeping track of offset & size info in the db as well, so that multiple blobs could be contained in the same file. Then the file gets split up when added as normal, but the store just points the different parts to different chunks of that same physical file.

As far as the problem of including the data in perkeep without storing it twice -- what I do here is to just dump the data into perkeep and then access it via the fuse filesystem.

Something like this would have been nice to have when I was doing a mass import of my backups months ago; not having to actually copy 2TiB of data might have sped things up a bit. I probably still would have had to write custom tooling though...

bobg:
> you could solve that by keeping track of offset & size info in the db as well, so that multiple blobs could be contained in the same file

Thanks for that suggestion! It's implemented in https://github.com/perkeep/perkeep/pull/1282/commits/4d9ca00d3f780f6ffc75ea0007de3d33b8386364. That commit also adds a new type, `FileSectionReader`, that, if used by `schema.WriteFileChunks`, would be most of what this needs to be fully useful. That's a project for a future commit on this branch.

bobg:
Just discovered there's an existing feature request for this: https://github.com/perkeep/perkeep/issues/1226.

tgulacsi (approved):

Issue #318, opened 2014-01-05
compress JSON responses


bradfitz:
<pre>JSON compresses well.

Do we even send compressed HTTP responses to e.g. search queries?</pre>

PR #1284, opened 2019-12-01
pkg/webserver: use NYTimes/gziphandler

bobg:
Fixes https://github.com/perkeep/perkeep/issues/318. Maybe overkill? This compresses all server responses; that issue suggested compressing only JSON responses.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898
 	cloud.google.com/go v0.39.0
 	github.com/FiloSottile/b2 v0.0.0-20170207175032-b197f7a2c317
+	github.com/NYTimes/gziphandler v1.1.1
 	github.com/aws/aws-sdk-go v1.14.31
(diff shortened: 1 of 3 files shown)


tgulacsi (approved):

aviau:
Do we really want to introduce a new outside dependency for something as simple as this?

tgulacsi:
You're right, but this 532 lines is almost the minimum we'd need to maintain ourselves, with its 673 lines of tests.
The parsing of Accept-Encoding is not trivial, and this simple-at-surface excercise has several non-trivial edge cases,
as I skimmed through that 532 lines.

Mohammed90:
Isn't the NYTimes package dead now, and the community is moving towards https://github.com/klauspost/compress/tree/master/gzhttp ?

tgulacsi:
I'm a big fan of @klauspost 's work, but that's an even bigger dependency - though faster, better.

aviau:
If the NYTimes is really unmaintained, then I think we can close this and someone can propose an alternative if needed.

PR #1288, opened 2019-12-12
all: make code "go vet"-clean

bobg:
Mostly this involved adding keys to composite literals, which `go vet` recently started requiring. However, there are a few legitimate bugfixes in here:

- Some structs were being copied with `a := *b` that should not have been, because they included `sync.Mutex`es.
- Some unreachable code was meant to be retrying indexing failures.
- Some context cancelers were not being called.

This PR also runs the whole tree through `go fmt`, producing a few additional diffs.


aviau:
I've rebased this PR and fixed conflicts in https://github.com/perkeep/perkeep/pull/1379. Closing this one for now.

Issue #878, opened 2016-11-05
deadlock between indexing & search (via importers)


bradfitz:
I keep getting camlistored deadlocks trying to reindex.

No web browser is open. (No websocket), which was a problem in the past.

What I see in the stacks is:

```
goroutine 2732318 [semacquire, 10 minutes]:
sync.runtime_Semacquire(0xc420a803a4)
        /home/bradfitz/go1.7/src/runtime/sema.go:47 +0x30
sync.(*RWMutex).RLock(0xc420a80398)
        /home/bradfitz/go1.7/src/sync/rwmutex.go:43 +0x5d
camlistore.org/pkg/index.(*Index).RLock(0xc420a80360)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/index/index.go:88 +0x31
camlistore.org/pkg/search.(*Handler).Query(0xc420200540, 0xc458246000, 0x0, 0x0, 0x0)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/search/query.go:867 +0x19c
camlistore.org/pkg/importer.(*importer).Node(0xc4206a13b0, 0x0, 0x0, 0x0)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/importer/importer.go:948 +0x32b
camlistore.org/pkg/importer.(*importer).credentials(0xc4206a13b0, 0xa1, 0xa1, 0xa1, 0x0, 0xa1, 0xa)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/importer/importer.go:792 +0x40
camlistore.org/pkg/importer.(*RunContext).Credentials(0xc45cbab500, 0xc455063000, 0xa1, 0x412498, 0xc480f37d60, 0x8, 0x0)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/importer/importer.go:331 +0x32
camlistore.org/pkg/importer/twitter.(*imp).Run(0x1555d60, 0xc45cbab500, 0x0, 0x0)
        /home/bradfitz/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/importer/twitter/twitter.go:167 +0x96
[...]

PR #1289, opened 2019-12-12
pkg/index: delay (even longer) acquiring a write lock in ReceiveBlob

bobg:
Possible fix for #878. Have been running with this change for a little while, haven't seen the deadlock again yet, pretty sure I would have by now. Also, `go test -race ./...` reports no (new) race conditions.

### DIFF
--- a/pkg/index/corpus_test.go
+++ b/pkg/index/corpus_test.go
@@ -30,7 +30,6 @@ import (
 	"perkeep.org/pkg/types/camtypes"
 )
 
-
 func newTestCorpusWithPermanode(t *testing.T) (c *index.Corpus, pn blob.Ref, keyID1, keyID2 string) {
 	c = index.ExpNewCorpus()
 	pn = blob.MustParse("abc-123")
@@ -113,7 +112,6 @@ func newTestCorpusWithPermanode(t *testing.T) (c *index.Corpus, pn blob.Ref, key
 
 	return c, pn, indextest.KeyID, keyID2
 }
-
 
 // TODO(mpl): remove that whole test?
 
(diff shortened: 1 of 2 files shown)


bradfitz:
It'd be nice to understand the problem more & ideally have a test for it, though.

bobg:
Agreed. The root cause is pretty clear: `Index.ReceiveBlob` was calling `populateFile` while holding the write lock, which sent requests (handled in another goroutine) that ended up also wanting to get the write lock in `Index.ReceiveBlob` when file-chunk blobs had to be added.

There are two things I still don't quite understand. First, why did this deadlock not happen every time? And second, how did it spontaneously resolve after a while? A timeout on the calling goroutine's http request could explain that, but the amount of time before resolving was too long and too irregular, and I never saw any error messages that said "timeout" or "context canceled" or anything like that.

bobg:
> The root cause is pretty clear

I take it back. I misunderstood `populateFile`, which is (in principle, at least) a read-only operation w.r.t. the blob server, though perhaps the concrete type of the `Fetcher` it takes does some mutation somewhere. The goroutine holding a write lock in [this example](https://github.com/perkeep/perkeep/issues/878#issuecomment-564889584) is _not_ waiting for some other perkeepd thread to satisfy a write request, it's just reading from Google Cloud Storage.

The other goroutine that's waiting for a write lock in that example is part of the sync loop, which seems fine.

Maybe there's something about the sync being preempted for a long time by lots and lots of writes that eventually causes a write to get stuck while holding the lock?

One thing I do know: the change in this PR definitely fixes the problem. I ran a `cp -a` that spent over a week populating my GCS perkeep storage and it didn't deadlock at all.

bradfitz:
Actually it seems like this makes the index tests flaky. It seems related to the random order that blobs are indexed and the inter-blob dependency tracking & rescheduling.

I'm debugging.

bobg:
Thanks. Let me know if I can help, although I won't be available to do so for about another week.

Regarding [this comment](https://github.com/perkeep/perkeep/pull/1289#issuecomment-567788111):
> It'd be nice to understand the problem more & ideally have a test for it, though.

I still agree.

PR #1290, opened 2019-12-14
make.go: add -offline flag

### LINKED ISSUE #1286: How to compile without internet access?
I'm regularly in a situation where I'm working without internet access. Unfortunately it seems that running `go run make.go` in this situation doesn't work because it still wants to contact a remote server. I've already compiled the code beforehand with internet access, so I already have all of the dependencies available locally. Is there a way to make this work?

mpl:
Also update to Go1.13.5

Fixes #1286


bradfitz:
Thanks!

PR #1291, opened 2019-12-23
Audio file info display and "media" predicate

mnzaki:
This PR adds some UI improvements to audio content blobs, and adds a "media" predicate that searches through all tags.

Some context:
I am hacking on the audio file handling a bit as I want to use perkeep to play music from my collection on my phone, and just wanted to share this early on. The next thing I wanted to do is have a central player element instead of a per-file one, so that I can navigate while it's playing.

### DIFF
--- a/pkg/search/predicate.go
+++ b/pkg/search/predicate.go
@@ -121,6 +121,11 @@ func init() {
 	registerKeyword(newIsPortait())
 	registerKeyword(newWidth())
 
+	// MediaTags predicates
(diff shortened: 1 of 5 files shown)


mnzaki:
@googlebot signed it

MichaHoffmann:
Is this PR still alive? I would also like to store my audio files on perkeep and find the UI elements and tag searchability a little lacking. Its working with raw queries, but a little cumbersome.

PR #1310, opened 2020-06-27
doc/readme: Updated Link

ChiragSukhala:
#Issue: 
The link pointing to "camlistore.org" will redirect back to "perkeep.org" anyways, so updated with relative link.
#Change:
Updated "Contributing" link from "https://camlistore.org/doc/contributing" to "/code#contributing".


mpl:
@willnorris Hi. no objection to this change?

PR #1312, opened 2020-07-04
Implement Google Takeout importer

NiklasMerz:
This is still work in progress. Opening this PR for early feedback and tracking progress.

## Background
Google offers you to export all your data in ZIP archives called Google Takeout. I am particular interested in data from Google Notes(Keep) since it offers still no API.

Planning to build this importer to import as much data as possible from the many services in the Google Takeout files.


NiklasMerz:
Importing notes works now. It's not pretty but it works for notes with just text content,

PR #1317, opened 2020-09-14
move continuous integration tests to GitHub Actions

willnorris:
This allows bringing both linux and windows CI tests into a single platform, and sharing accounts/permissions with the rest of the repo rather than separated out on Travis and AppVeyor.  If we also use GitHub Actions for future automations like release scripts, updating the website, etc, then having CI here as well means only a single system to learn.

Performance is about the same as Travis and AppVeyor... about 5-6 minutes for linux tests to complete, about 8 minute for Windows.  The failing Windows tests here are the same ones that were already failing.  I have a set of changes queued up that will address all of those.


willnorris:
Example test results can be seen at https://github.com/willnorris/perkeep/actions

bradfitz:
Hell yeah, thanks!

.github/workflows/tests-linux.yml
```diff
@@ -0,0 +1,32 @@
+on: [push, pull_request]
+name: "tests/linux"
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.x, 1.13.x]
```
bradfitz, line 8: Don't we require Go 1.14 at least?

bradfitz (approved):

.github/workflows/tests-linux.yml
willnorris: still says 13 here: https://github.com/perkeep/perkeep/blob/7321b90b6e112e7847cb6f3f9683c74d545d5059/make.go#L901

mpl (commented):
Is there zero advantage in keeping the Travis runs around as well? Say for when github craps itself for example?

willnorris:
I don't really think we need to keep it around.  And if GitHub really does fall down in some terrible way, we can always bring back the travis config.

mpl (approved):
I don't know the github actions, trusting you on that one. :-)

PR #1318, opened 2020-09-15
Fix CI tests for windows

willnorris:
This also includes the changes from #1317, just so I can test GitHub Actions.  Once, #1317 is merged, I'll rebase this PR.

Test runs: 

[tests/linux](https://github.com/willnorris/perkeep/actions?query=workflow%3Atests%2Flinux+branch%3Afix-windows) 
[tests/windows](https://github.com/willnorris/perkeep/actions?query=workflow%3Atests%2Fwindows+branch%3Afix-windows)


cmd/pk-put/kvcache.go
```diff
@@ -346,7 +346,7 @@ func fileInfoToFingerprint(fi os.FileInfo) statFingerprint {
 		if clean := cleanSysStat; clean != nil {
 			// TODO: don't clean bad fields, but provide a
 			// portable way to extract all good fields.
-			// This is a Linux+Mac-specific hack for now.
+			// This is a Linux+Mac+Windows-specific hack for now.
```
mpl, line 349: At which point does it become faster to list the platforms for which the hack does not exist? ;)

willnorris: sure.  Do we have an official list of platforms that Perkeep supports?  Because we could also just work on adding support for those platforms.

willnorris:
rebased, and should be ready to go once CI tests finish running.

cmd/pk-put/kvcache.go
bradfitz: Linux, Mac, Windows, FreeBSD probably.

mpl:
@bradfitz I want to merge this, but @willnorris has (understandably) not adjusted the vendoring accordingly to the changes in go.mod, because it is actually unmaintainable with 'go mod vendoring' without breaking gopherjs.
I've just tried doing it by hand, but it is horrible, and this time I haven't even succeeded (somehow now vendor/golang.org/x/tools seems to be needed, even though it does not seem we had it before. beats me).
I vote that we ditch GopherJS and give up on being able to rebuild the JS part (i.e. the -buildWebUI of make.go would be gone) for now? What do you all think?

cmd/pk-put/kvcache.go
willnorris: apparently, this seems to work fine on FreeBSD already (tested on FreeBSD 12.1), so I've updated this commit to just remove this "hack" comment entirely.  I've left the TODO in place however, since it does still seem like a good idea to extract good fields rather than clean bad fields.

willnorris:
@mpl are you suggesting dropping gopherjs entirely?  What would that entail?  I haven't been involved in any of the frontend development.  Does that basically involve rewriting the frontend code in vanilla JS?  How much is there?  (Though I guess according to https://github.com/perkeep/perkeep/commit/f127d47e it looks like you were already moving away from GopherJS anyway)

mpl:
> @mpl are you suggesting dropping gopherjs entirely? What would that entail? I haven't been involved in any of the frontend development. Does that basically involve rewriting the frontend code in vanilla JS? How much is there? (Though I guess according to [f127d47](https://github.com/perkeep/perkeep/commit/f127d47e) it looks like you were already moving away from GopherJS anyway)

(already answered by chat, but for others)
Yes, dropping it completely. A lot of the code base is still vanilla JS, bits of React, etc. But a few features are generated by GopherJS. As soon as we give up on maintaining the vendoring for the GopherJS toolchain to work, it means we at least lose the ability to regenerate (and therefore modify) the JS that is behind these features.

bradfitz:
We're going to drop gopherjs and move to just JavaScript + Go 1.15 (or higher) + Go modules (dropping vendor/), etc. Plan is:

https://github.com/perkeep/perkeep/pull/1349#issuecomment-758952372

Then we can merge this it sounds like.

/cc @aviau

mpl:
> We're going to drop gopherjs and move to just JavaScript + Go 1.15 (or higher) + Go modules (dropping vendor/), etc. Plan is:
> 
> [#1349 (comment)](https://github.com/perkeep/perkeep/pull/1349#issuecomment-758952372)
> 
> Then we can merge this it sounds like.
> 
> /cc @aviau

Yes, this PR is only blocked because it doesn't keep the vendor dir synced correctly wrt to gopherjs. If we drop both then no problem. We've actually already drifted anyway, with a PR that got merged a few weeks (?) ago and that did not update the vendor dir correspondingly.

aviau:
The goplusjs PR was merged. It seems @mpl reviewed and wanted to merge this so I will go ahead and do it.

aviau:
I had to disable the check from b2983caad461930d8d35e1fddaefa511a3a70cd6 on Windows. Looks like we can check that a file is readonly but not by who it is readable using the permissions bits. Let me know if you think of a relevant check to perform, in the meantime we have a working windows build :tada:

PR #1323, opened 2020-10-30
all: update to Go 1.15

mpl:


Dockerfile
```diff
@@ -22,11 +22,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		pkg-config \
                 libsqlite3-dev
 
-ENV GOLANG_VERSION 1.13.5
+ENV GOLANG_VERSION 1.15.3
```
bradfitz, line 25: nice transposition :)

make.go
```diff
@@ -898,7 +898,7 @@ func validateDirInGOPATH(dir string) error {
 }
 
 const (
-	goVersionMinor  = 13
+	goVersionMinor  = 15
 	gopherJSGoMinor = 12
```
bradfitz, line 902: gopherjs still a problem for us? can we nuke it and move to wasm yet?

Dockerfile
mpl: ;)

PR #1326, opened 2020-12-05
go.mod: update "x/crypto" to support latest acme API

NiklasMerz:
Let's encrypt doesn't work with the old acme API anymore. The latest version of "x/crypto" fixes that.

Error messages:
```
acme/autocert: missing certificate
403 urn:acme:error:unauthorized: Error creating new authz :: Validations for new domains are disabled in the V1 API (https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430)
```

See:https://go-review.googlesource.com/c/crypto/+/200638

I ran `go get -u golang.org/x/crypto` to update the module. Tests pass with devcam and my server got Let's Encrypt working again.


bradfitz:
Thanks!

mpl:
I notice this PR got merged even though it does not update the vendor directory. Are we giving up on keeping the go.mod and the vendor directory in sync?

PR #1332, opened 2021-01-02
serverinit: two-way sync by default

### LINKED ISSUE #1331: s3 + blobPath should setup two-way sync
Hello!

I was expecting that configuring both s3 and blobPath would setup two-way sync. However running ``pk sync --all`` will only send the blobPath blobs to s3 without pulling new ones from s3.

Is there a reason for this? Would there be downsides to enabling two-way sync by default?

It seems like this is acheivable with the low-level config but I was under the impression that it would be a good default for the high-level one.

aviau:
closes #1331

This makes setting up perkeep on a new computer pretty easy. Paste your config and sync.

I wonder why this wasn't the default, could it be because listing remote objects can take a long time?

### DIFF
--- a/pkg/serverinit/genconfig.go
+++ b/pkg/serverinit/genconfig.go
@@ -570,6 +570,19 @@ func (b *lowBuilder) addS3Config(s3 string) error {
 					"file": filepath.Join(b.high.BlobPath, "sync-to-s3-queue."+b.kvFileType()),
 				}),
 		})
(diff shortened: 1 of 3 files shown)


bradfitz (commented):
Let's drop this for now but when we do a new config format, we'll require an explicit direction either way: either `oneWay` or `bidirectional` or something.

aviau:
Yep, closing.

PR #1333, opened 2021-01-03
docs: s3:DeleteObject is not required

aviau:
AFAIK this is not needed unless there are things in perkeep that deletes objects?


aviau:
Ah, blobpacked needs it!

PR #1334, opened 2021-01-03
genconfig: support blobpacked in cloud replicas

### LINKED ISSUE #1330: help on setting-up s3-only instance
Hello!

First, I want to thank you for perkeep! This is an amazing project and I intend to send pull requests to improve both the code and documentation.

I'd like for clarifications on how to setup an s3-only instance.

I have successfully managed to setup a blobPath+s3 instance, which works well:
```
[...]
    "packRelated": true,
    "blobPath": "/home/aviau/var/perkeep/blobs",
    "s3": "<censored>",
[...]
```

The documentation says:
> At least one of these must be set

Leading me to think that s3-only should work.

Removing blobPath causes the UI to still show the files, but clicking any of them throws errors:
```
Can't serve file: could not fetch sha224-b3741359a37aeebd5af04a5d539dca7546245362e6b10d41e681cbbf: file does not exist
```

And ``pk list`` shows none of my blobs.

Then I tought that maybe I need to reindex, so I did that, and it only caused my files to disapear from the UI.

Is there something I am not doing correctly?

aviau:
closes #1330

    Cloud-backed storage currently does not use blobpacked
    if they are a replica. This causes issues if they become
    the main storage.
    
    Example use case where this would cause issues:
     - Create a configuration with s3, blobPath, and packRelated.
     - Add objects
     - Remove blobPath
     - `pk list`
     - All blobs are gone because the s3-only config uses blobpacked
       but the s3 replica was not.

Let me know if there is something I overlooked! I think this reduces friction with using the simple configuration.

### DIFF
--- a/pkg/serverinit/genconfig.go
+++ b/pkg/serverinit/genconfig.go
@@ -546,18 +546,51 @@ func (b *lowBuilder) addS3Config(s3 string) error {
 		hostname = f[3]
 	}
 	isReplica := b.hasPrefix("/bs/")
-	s3Prefix := ""
-	s3Args := args{
-		"aws_access_key":        accessKey,
-		"aws_secret_access_key": secret,
-		"bucket":                bucket,
-	}
-	if hostname != "" {
-		s3Args["hostname"] = hostname
-	}
+	s3Prefix := "/bs/"
 	if isReplica {
 		s3Prefix = "/sto-s3/"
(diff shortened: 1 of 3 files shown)


bradfitz (approved):
I'll let you decide whether you want to submit this.

I still want to redo the whole config system sometime.

aviau:
Sure, I am looking forward to a nicer config system too.

I'll still try to support the current one in the meantime.

PR #1335, opened 2021-01-03
server-config: packRelated doesn't need blobPath

aviau:
AFAIK this was false. `packRelated` affects `s3`, `b2` and `googlecloudstorage`.

It isn't supported on some other storage types like `memoryStorage` or  `googledrive` but it will raise errors so the user will know.


bradfitz (approved):
If this is true. I forget. :)

PR #1336, opened 2021-01-03
b2: document example + point to perkeep/b2 fork

aviau:
- Add documentation in storage-examples for using Perkeep with b2 as a
  storage backend. The documentation uses the new API terminology:
  AccountID is no longer a thing in the Backblaze web console.

- Replace FiloSottile/b2 by perkeep/b2 where we host a copy of
  clgillis's Pull Request[1]. We can point to the original project
  again as soon as it is merged.

  Note that I haven't updated the vendor directory and it seems that
 `go mod vendor` updates way more things that are also outdated in the
  master branch, so I will leave that as an exercise to whoever has
  commit access :).

Before this Pull Request, launching the b2 storage would error with the
following error message:
```
Error parsing config: Caught panic: error instantiating storage for prefix "/sto-b2/", type "b2": Post "https://api000.backblazeb2.com/b2api/v1/b2_list_buckets": b2 remote error [bad_request]: accountId invalid
```

### DIFF
--- a/doc/storage-examples.md
+++ b/doc/storage-examples.md
@@ -6,17 +6,17 @@
 
 To use S3 with Perkeep, you need to:
 
-* Sign up for an Amazon Web Services account
-* Sign into the AWS console, navigate to 'S3'
-* Create an S3 bucket for your Perkeep backups (Perkeep will not work if you put other files in this bucket)
-* Configure your Perkeep server to sync blobs to the S3 bucket
+* Sign up for an Amazon Web Services account.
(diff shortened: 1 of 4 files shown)


bradfitz (approved):

aviau:
@bradfitz 

I have removed the "replace" from go.mod and I moved my fork to perkeep/b2 instead. Is that fine with you?

cancercare-machine-user (approved):
can connect to b2 using app key

bradfitz (approved):

PR #1339, opened 2021-01-04
bytes.md: part must contain BlobRef or BytesRef

aviau:
I confirmed this by looking at schema.go:

https://github.com/perkeep/perkeep/blob/4b2ae528420fe68a2b95d7d5fc7e3f4490c4d91e/pkg/schema/schema.go#L772

I was wondering what a size-only part could possibly mean.

### DIFF
--- a/doc/schema/bytes.md
+++ b/doc/schema/bytes.md
@@ -27,6 +27,9 @@ a "file" schema blob.
       //                these in a hash tree.  it is an error if both "bytesRef"
       //                and "blobRef" are specified.
       //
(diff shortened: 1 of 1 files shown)


bradfitz:
This change isn't right. A chunk with a size and no data is how you represent a hole of a sparse file. Or just zeros.

aviau:
Ahh, right. I have modified the PR to document this, as it wasn't immediately clear for me when I read the docs.

bradfitz (approved):

PR #1341, opened 2021-01-05
cmd/pk/list: filter by type

aviau:
filter blobs by schema type. I piped to grep a couple time but this is a bit better.

Also changed `blob type` for `schema blob type` because that's what we call them according to [docs/terms](https://perkeep.org/doc/terms).

### DIFF
--- a/cmd/pk/list.go
+++ b/cmd/pk/list.go
@@ -34,8 +34,9 @@ import (
 type listCmd struct {
 	*syncCmd
 
-	describe bool           // whether to describe each blob.
-	cl       *client.Client // client used for the describe requests.
+	describe  bool           // whether to describe each blob.
+	camliType string         // filter by schema blob type
+	cl        *client.Client // client used for the describe requests.
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/pk/list.go
```diff
@@ -45,9 +46,11 @@ func init() {
 				dest: "stdout",
 			},
 			describe: false,
+			camliType: "",
 		}
 		flags.StringVar(&cmd.syncCmd.src, "src", "", "Source blobserver is either a URL prefix (with optional path), a host[:port], a path (starting with /, ./, or ../), or blank to use the Perkeep client config's default host.")
-		flags.BoolVar(&cmd.describe, "describe", false, "Use describe requests to get each blob's type. Requires a source server with a search endpoint. Mostly used for demos. Requires many extra round-trips to the server currently.")
+		flags.BoolVar(&cmd.describe, "describe", false, "Use describe requests to get each schema blob's type. Requires a source server with a search endpoint. Mostly used for demos. Requires many extra round-trips to the server currently.")
+		flags.StringVar(&cmd.camliType, "type", "", "Filter by schema blob type. Implies -describe.")
```
bradfitz, line 53: Empty means no filter.

bradfitz (approved):

PR #1343, opened 2021-01-07
make: Don't strictly require `go1.12`

lenormf:
The [gopherjs documentation](https://github.com/gopherjs/gopherjs/blob/master/README.md) states:

        GopherJS requires Go 1.12 or newer.

This commit prevent offline builds from erroring out when using a recent version of `go`.

### DIFF
--- a/make.go
+++ b/make.go
@@ -943,7 +943,7 @@ func verifyGoVersion() {
 		return
 	}
 
-	if minorVersion != gopherJSGoMinor {
+	if minorVersion < gopherJSGoMinor {
 		verifyGopherjsGoroot(fmt.Sprintf("1.%d", minorVersion))
 	}
 }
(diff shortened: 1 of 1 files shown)


bradfitz:
GopherJS is a thorn in our project's side. It's been neglected and we need to migrate off of it. Until we do, we can't just pick any Go version to use with GopherJS... we have to use only a version of Go that GopherJS is compatible with.

To do a Perkeep build, I think we require two version of Go currently. @mpl, is that right?

lenormf:
I checked the issue tracker for GopherJS and [it does look like it only runs on `v1.12` for now](https://github.com/gopherjs/gopherjs/issues/963), which contradicts the bits in the documentation that I linked above.

In the end I built Perkeep in “online mode” because even when skipping this version check to accommodate GopherJS, `make.go` complained that the code had to be located in `$GOPATH`, which isn't possible when [building a package](https://aur.archlinux.org/packages/perkeep/) (ArchLinux package I made to build on ARM).

lenormf:
Off-topic question, since I don't want to create another irrelevant issue: do you have any plans for Perkeep (short/long term)? I see that @mpl stopped working on it, and neither them nor you are on the IRC channel.

mpl:
> GopherJS is a thorn in our project's side. It's been neglected and we need to migrate off of it. Until we do, we can't just pick any Go version to use with GopherJS... we have to use only a version of Go that GopherJS is compatible with.
> 
> To do a Perkeep build, I think we require two version of Go currently. @mpl, is that right?

Yes, to do a full offline build, we probably need the specific old version for gopherjs (Go 1.12?), and then for the actual Perkeep build the latest version.

aviau:
Note that we may be able to simplify the gopherjs build system by migrating to https://github.com/goplusjs/gopherjs which support 1.15.

mpl:
oh, that is interesting, thanks.

PR #1348, opened 2021-01-11
downloadbutton: migrate from goui to JS

aviau:
My understanding is that we want to get rid of gopherJS (with reason -- this is making perkeep unreasonable to build). So here is a port of the download button to Javascript.

I have tested:
 - single-file download
 - multi-file-zip download
 - building the goui to make sure it still builds (with the complicated go1.12 + go1.15 setup)

It works exactly the same as it did with the goui. I'll do more in exchange for code reviews.


bradfitz:
What's the endgame here? Rewrite all our Go UI in JavaScript? That doesn't sound super exciting. I thought we'd migrate to something WebAssembly-based instead (https://github.com/vugu/vugu or whatnot).

Do we still need this change if we do #1349? 

/cc @mpl

aviau:
> What's the endgame here? Rewrite all our Go UI in JavaScript? That doesn't sound super exciting.

Right. Not exciting but will work even long term.

> Do we still need this change if we do #1349?

Technically no, but I wouldn't expect gopherjs to be a long term good solution. While there seems to be energy behind the goplusjs fork I don't think that there will be any significant community forming behind it due to the fact that compiling to webassembly seems to be the future for Go.

However even building to webassembly is considered experimental and I wouldn't depend on any of those technologies just yet. Plain javascript will to just fine for the simple use cases in the webui.

I see goplusjs/gopherjs as a solution to keep our build system sane for the moment, but we should still migrate away from it.

aviau:
AFAIK @mpl agreed with that:

https://github.com/perkeep/perkeep/pull/1318#issuecomment-698656717

mpl:
> > What's the endgame here? Rewrite all our Go UI in JavaScript? That doesn't sound super exciting.
> 
> Right. Not exciting but will work even long term.
> 
> > Do we still need this change if we do #1349?
> 
> Technically no, but I wouldn't expect gopherjs to be a long term good solution. While there seems to be energy behind the goplusjs fork I don't think that there will be any significant community forming behind it due to the fact that compiling to webassembly seems to be the future for Go.
> 
> However even building to webassembly is considered experimental and I wouldn't depend on any of those technologies just yet. Plain javascript will work just fine for the simple use cases in the webui.
> 
> I see goplusjs/gopherjs as a solution to keep our build system sane for the moment, but we should still migrate away from it.

I don't know what our next framework will/should be, and I agree that having goplusjs instead of gopherjs in the meantime seems like a good solution to at least make things buildable and bearable again. And indeed I don't think we should build anything else/new on top of goplusjs.

However, until we know what our next step is, if things are working with goplusjs, I wouldn't rewrite them in vanilla js. Especially since I'm sure hoping vanilla js is not the next "framework" we use. But that's just me.

bradfitz (approved):

PR #1347, opened 2021-01-10
encrypt.go: check keyFile permissions

aviau:
picked-up a small todo in the code.

PR #1349, opened 2021-01-11
move to goplusjs

aviau:
gopherjs -> goplusjs
    
    github.com/gopherjs/gopherjs no longer supports new Go versions and
    doing a full perkeep build currently requires go 1.12.
    
    This moves to github.com/goplusjs/gopherjs which supports new Go
    version. Building perkeep no longer requires go 1.12.
    
    - support building webui with just go 1.15
    - `go mod tidy` produces no output anymore
    - `go mod vendor` is up to date
    - Add CI check for tidyness: nobody can break vendoring in master anymore.
    - goplusjs/gopherjs builds with gomodules

Its really annoying to have to run go mod commands and then add-back the stuff that was removed. I have worked around this with a `uideps` package. I haven't made much research on better ways to do this but it seems to work fine.

I haven't really checked for all details because I am not sure that there is interest to merge this, but AFAIK this is way better than the current state of things.


aviau:
maybe @mpl will be interested in reviewing this one?

mpl:
yeah, I don't know when I can get around to it, but I think this is one of our most urgent problems to solve.

aviau:
Cool. It still breaks tests but I'll fix it.

aviau:
Done. Tests fixed.

aviau:
I have split the PR into two commits so that the review is easier.

The first commit includes all important chances. The second commit only oncludes diffs to `vendor/`.

bradfitz:
Can you update the PR & commit message to explain the background on this more?

I hadn't even heard of goplusjs until now.

I assume the reason for this PR is that GopherJS is old & unmaintained and doesn't support new versions of Go. Does goplusjs support Go 1.15?

aviau:
> Can you update the PR & commit message to explain the background on this more?

Sure, done.

> I assume the reason for this PR is that GopherJS is old & unmaintained and doesn't support new versions of Go. Does goplusjs support Go 1.15?

Yes and Yes.

you can run `go run make.go -buildWebUI` with Go 1.15 now.

bradfitz:
Let's:

* require Go 1.15
* require Go modules; enforce that GO111MODULE mode is on (fail if it's set to `"off"`)
* nuke the whole vendor directory
* run the goplusjs binary as `go run ...../cmd/goplusjs` (it should be in the `go.mod`)

Then eventually, once all the GopherJS stuff is converted to goplusjs, then we can drop goplusjs entirely.

aviau:
Here we go. This removes 1.2M lines of code from the repository.

-**Everything** is built with gomodules
- Gopherjs is ran using `go run`
- reactGen is still ran using `reactGen`, but build with gomodules instead of  from `vendor/`

aviau:
cc @bradfitz for a second round of review

(also see https://github.com/perkeep/perkeep/pull/1347#issuecomment-759088259 when you have time, or I can't merge)

bradfitz (approved):
YAY!! Thanks!

PR #1352, opened 2021-01-14
downloadbutton: missing return + leftover file

aviau:
downloadbutton: missing return + leftover file
    
    Remove leftover generated file from
    1700b5be5f324d65913c7b7718dc06ffa2da42b9.
    
    Also add missing return, causing two downloads to start
    when selecting a single file.

This is the return that used to be here: https://github.com/perkeep/perkeep/blob/39bc87d2738ddf3885f1007480a4d258d49f9ce9/server/perkeepd/ui/goui/downloadbutton/downloadbutton.go#L141


aviau:
cc @bradfitz for review

aviau:
I am really confident about this, so I'll merge it, but let me know if you prefer that I wait even for simple stuff like this in the future.

PR #1353, opened 2021-01-14
CI: test on macOS

aviau:
Now we have a green build on linux, macos and windows :tada:


aviau:
cc @bradfitz for review!

bradfitz (approved):

PR #1354, opened 2021-01-16
aboutdialog: migrate from goui to JS

aviau:
Nothing changes, except for `s/Camlistore/Perkeep/`

# Before

Screenshot from 2021-01-15 22-33-37

# After

Screenshot from 2021-01-15 22-32-49


aviau:
cc @bradfitz for review :)

One more component migrated away from gopherjs.

bradfitz (approved):

bradfitz:
For future components you might want to dig into the git history. These might've all bee JavaScript already in the past before GopherJS was introduced.

aviau:
Yeah I was thinking this too when I saw the methods in ServerConection. I was surprised it implemented server status, which suggests it might have been in JS before.

PR #1356, opened 2021-01-17
pkg/schema: create CamliType type

aviau:
pkg/schema: create CamliType type
    
    Create a CamliType type in pkg/schema and use it in a couple of
    packages.
    
    It can be implemented in other packages as we go.

I was reading pkg/index and decided to implement @mpl's TODO:
```
// TODO(mpl): create consts somewhere for "claim" and "permanode" as camliTypes, and use them,		if meta.CamliType != string(schema.PermanodeCamliType) && meta.CamliType != string(schema.ClaimCamliType) {
// instead of hardcoding. Unless they already exist ? (didn't find them).
```

### DIFF
--- a/pkg/index/corpus.go
+++ b/pkg/index/corpus.go
@@ -69,7 +69,7 @@ type Corpus struct {
 
 	// camBlobs maps from camliType ("file") to blobref to the meta.
 	// The value is the same one in blobs.
-	camBlobs map[string]map[blob.Ref]*camtypes.BlobMeta
+	camBlobs map[schema.CamliType]map[blob.Ref]*camtypes.BlobMeta
 
 	// TODO: add GoLLRB to vendor; keep sorted BlobMeta
 	keyId signerFromBlobrefMap
@@ -326,7 +326,7 @@ func (pm *PermanodeMeta) valuesAtSigner(at time.Time,
 func newCorpus() *Corpus {
 	c := &Corpus{
 		blobs:                   make(map[blob.Ref]*camtypes.BlobMeta),
-		camBlobs:                make(map[string]map[blob.Ref]*camtypes.BlobMeta),
+		camBlobs:                make(map[schema.CamliType]map[blob.Ref]*camtypes.BlobMeta),
 		files:                   make(map[blob.Ref]camtypes.FileInfo),
 		permanodes:              make(map[blob.Ref]*PermanodeMeta),
 		imageInfo:               make(map[blob.Ref]camtypes.ImageInfo),
@@ -463,8 +463,8 @@ func (c *Corpus) scanFromStorage(s sorted.KeyValue) error {
 		return err
 	}
 
-	c.files = make(map[blob.Ref]camtypes.FileInfo, len(c.camBlobs["file"]))
-	c.permanodes = make(map[blob.Ref]*PermanodeMeta, len(c.camBlobs["permanode"]))
+	c.files = make(map[blob.Ref]camtypes.FileInfo, len(c.camBlobs[schema.TypeFile]))
+	c.permanodes = make(map[blob.Ref]*PermanodeMeta, len(c.camBlobs[schema.TypePermanode]))
 	cpu0 := osutil.CPUUsage()
 
 	var grp syncutil.Group
@@ -675,7 +675,7 @@ func (c *Corpus) mergeBlobMeta(bm camtypes.BlobMeta) error {
 	if _, dup := c.blobs[bm.Ref]; dup {
 		panic("dup blob seen")
 	}
-	bm.CamliType = c.str(bm.CamliType)
+	bm.CamliType = schema.CamliType((c.str(string(bm.CamliType))))
 
 	c.blobs[bm.Ref] = &bm
 	c.sumBlobBytes += int64(bm.Size)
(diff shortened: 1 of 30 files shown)


pkg/schema/schema.go
```diff
@@ -785,9 +785,27 @@ func populateParts(m map[string]interface{}, size int64, parts []BytesPart) erro
...
 func newBytes() *Builder {
-	return base(1, "bytes")
+	return base(1, BytesCamliType)
 }
 
+// CamliType is one of the valid "camliType" fields in a schema blob. See doc/schema
+type CamliType string
+
+const (
+	BytesCamliType     CamliType = "bytes"
```
bradfitz, line 795: How about "TypeBytes", "TypeClaim", etc, dropping "Camli" entirely from at least the const name?

pkg/search/handler.go
```diff
@@ -808,14 +809,14 @@ func (h *Handler) EdgesTo(req *EdgesRequest) (*EdgesResponse, error) {
 		if found {
 			ei = &EdgeItem{
 				From:     edge.From,
-				FromType: "permanode",
+				FromType: string(schema.PermanodeCamliType),
```
bradfitz, line 812: Why not change the EdgeItem.FromType type?

pkg/search/query.go
```diff
@@ -383,7 +384,7 @@ func (c *Constraint) matchesAtMostOneBlob() blob.Ref {
 }
 
 func (c *Constraint) onlyMatchesPermanode() bool {
-	if c.Permanode != nil || c.CamliType == "permanode" {
+	if c.Permanode != nil || c.CamliType == string(schema.PermanodeCamliType) {
```
bradfitz, line 387: likewise, why not just the CamliType type?

pkg/index/corpus.go
```diff
@@ -463,8 +463,8 @@ func (c *Corpus) scanFromStorage(s sorted.KeyValue) error {
 		return err
 	}
 
-	c.files = make(map[blob.Ref]camtypes.FileInfo, len(c.camBlobs["file"]))
-	c.permanodes = make(map[blob.Ref]*PermanodeMeta, len(c.camBlobs["permanode"]))
+	c.files = make(map[blob.Ref]camtypes.FileInfo, len(c.camBlobs[string(schema.FileCamliType)]))
```
bradfitz, line 466: likewise: camBlobs could be keyed by the new type?

bradfitz (commented):
There are a lot of `string(foo)` in this change, which suggests something isn't fully right yet. See comments.

pkg/schema/schema.go
aviau: Looks good, done.

pkg/search/handler.go
aviau: Done

pkg/search/query.go
aviau: Done

pkg/index/corpus.go
aviau: Done

aviau:
I went the soft route at first, transforming it to strings all the time. However you are right we should just use CamliType everywhere.

I just pushed a new version of the patch that does just that. I also renamed it as you suggested.

aviau:
@bradfitz Alright, done!

- `string` -> `CamliType` everywhere it made sense
- Renamed types to `Type<...>` as you suggested.

pkg/index/corpus.go
```diff
@@ -675,7 +675,7 @@ func (c *Corpus) mergeBlobMeta(bm camtypes.BlobMeta) error {
 	if _, dup := c.blobs[bm.Ref]; dup {
 		panic("dup blob seen")
 	}
-	bm.CamliType = c.str(bm.CamliType)
+	bm.CamliType = schema.CamliType((c.str(string(bm.CamliType))))
```
aviau, line 678: Not really sure if that works, can you confirm?

bradfitz: converting between two string-backed types is free. they'll share the same memory.

pkg/server/download.go
```diff
@@ -416,7 +416,7 @@ func (dh *DownloadHandler) statFiles(refs []blob.Ref) error {
 	return nil
 }
 
-var allowedFileTypes = map[string]bool{"file": true, "symlink": true, "fifo": true, "socket": true}
+var allowedFileTypes = map[schema.CamliType]bool{schema.TypeFile: true, schema.TypeSymlink: true, schema.TypeFIFO: true, schema.TypeSocket: true}
```
bradfitz, line 419: I'd break this up onto 6 lines

pkg/schema/schema.go
```diff
@@ -785,9 +785,27 @@ func populateParts(m map[string]interface{}, size int64, parts []BytesPart) erro
 }
 
 func newBytes() *Builder {
-	return base(1, "bytes")
+	return base(1, TypeBytes)
 }
 
+// CamliType is one of the valid "camliType" fields in a schema blob. See doc/schema
```
bradfitz, line 791: end in period.

bradfitz (approved):

pkg/server/download.go
aviau: Yeah, much nicer.

pkg/schema/schema.go
aviau: done!

PR #1358, opened 2021-01-18
handleAbout: move to index.js

aviau:
handleAbout: move to index.js
    
    All other buttons from the top menu are implemented like this, and
    the go ones were an exception. This isn't in go anymore so we should
    just use `getMenuItem_` like all others and implement the event
    handler in index.js.


aviau:
cc @bradfitz for review :)

bradfitz (approved):

PR #1363, opened 2021-01-19
pk-put+perkeepd config: support fingerprints

### LINKED ISSUE #1361: pk-put init --gpgkey does not accept fingerprint
Only works with short keyid.

Also, the error when you run init with no switches tells you there is a switch `--secret-keyring` ... there is not.  `CAMLI_SECRET_RING` env var works, though.

aviau:
closes #1361

```
    pk-put+perkeepd config: support fingerprints
    
    - Key ids in the `0xXXXXXXXXXXXXXXXX` format are now accepted.
    
    - Key ids in the `02A6 ACA0 5F16 9329 D94A  FFB8 C804 D0BA 9F70 883D`
      format are also accepted. With or without spaces.
    
    - Help texts now use long keyids so that its clear that we support
      them.
```

### DIFF
--- a/cmd/pk-put/init.go
+++ b/cmd/pk-put/init.go
@@ -87,7 +87,7 @@ func (c *initCmd) Examples() []string {
 	// like a lame hack.
 	return []string{
 		"",
-		"--gpgkey=XXXXX",
+		"--gpgkey=XXXXXXXXXXXXXXXX",
 		"--newkey #Creates a new identity",
 	}
 }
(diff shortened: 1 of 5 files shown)


aviau:
@bradfitz review plz!

aviau:
WIP. I'll modify this PR further so that we can accept key fingerprints.

aviau:
Done. Fingerprints are now supported.

bradfitz (approved):

PR #1364, opened 2021-01-20
signer: deprecate KeyId()

aviau:
Noticed this while working on https://github.com/perkeep/perkeep/pull/1363.

I think we should use fingerprints as much as possible and revert to shorter ids only when necessary, possibly fully deprecating short ids if that is possible. This PR is one step towards this. I intend to open a couple of follow-up PRs improving on this, propagating fingerprints where it makes sense and using them as much as possible.

```
    signer: deprecate KeyId()
    
    `signer.KeyId()` was only used in a counter-productive way
    by comparing it with half of a long ID.
    
    Also use this opportunity to return the full fingerprint from
    ParseArmoredPublicKey. NewSigner was the only caller of this
    function.
```

### DIFF
--- a/pkg/schema/sign.go
+++ b/pkg/schema/sign.go
@@ -35,7 +35,6 @@ import (
 // A Signer signs the JSON schema blobs that require signing, such as claims
 // and permanodes.
 type Signer struct {
-	keyID      string // short one; 8 capital hex digits
 	pubref     blob.Ref
 	privEntity *openpgp.Entity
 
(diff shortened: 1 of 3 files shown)


aviau:
cc @bradfitz for review!

pkg/schema/sign.go
```diff
@@ -81,7 +75,7 @@ func NewSigner(pubKeyRef blob.Ref, armoredPubKey io.Reader, privateKeySource int
 	case *openpgp.Entity:
 		privateKey = v
 	case string:
-		privateKey, err = jsonsign.EntityFromSecring(keyID, v)
+		privateKey, err = jsonsign.EntityFromSecring(fingerprint[32:], v)
```
bradfitz, line 78: what's the intention of this `[32:]` here? not obvious.

aviau: This PR was based on the master branch before  https://github.com/perkeep/perkeep/pull/1363 was merged. Back then, `EntityFromSecring` would not accept a fingerprint, so I took the last 8 from the fingerprint to get the short id.

However now that https://github.com/perkeep/perkeep/pull/1363 is merged we can pass the full fingerprint, which was my master plan from the start ;)

aviau:
cc @bradfitz for a second round of review!

bradfitz (approved):

PR #1366, opened 2021-01-21
IsBlobRef: go -> JS

aviau:
Almost done. Only the geo thigs are left and there won't be any go in the main UI.


aviau:
cc @bradfitz for review!

bradfitz (approved):

PR #1367, opened 2021-01-21
mapquery: Go -> JS

aviau:
The last one!

When combined with the other pending PRs, this completely removes gopherjs from the main UI.

There is still some gopherjs left in the publisher but really not that much.


aviau:
cc @bradfitz for review!

bradfitz (approved):

PR #1368, opened 2021-01-21
buildsystem: deprecate perkeepd goui

aviau:
buildsystem: deprecate perkeepd goui
    
    - GopherJS is no longer required to build the perkeepd ui.
    
    - Rename the `-buildWebUI` flag to `-buildPublisherUI`. Only the
      publisher needs gopherjs.
    
    - Adapt `make.go` so that it no longer tries to generate the perkeepd
      web interface with GopherJS.
    
    - Adapt `make.go` so that it no longer tries to download goui.js from
      perkeep.org as it isn't needed anymore.


aviau:
cc @bradfitz for review!

bradfitz (approved):

aviau:
I haven't timed it but I am noticing faster load times. It makes sense because the goui was embeding a ton of stuff from the go stdlib.

PR #1369, opened 2021-01-22
delete.go: implement Examples()

aviau:
So that `delete` is listed under `pk put -help` examples.


aviau:
cc @bradfitz for review

bradfitz (approved):

PR #1370, opened 2021-01-22
haveImageID: `-f` -> `--format`

aviau:
haveImageID: `-f` -> `--format`
    
    Ethier this has never worked or the Docker CLI's API has changed. The
    current meaning of the `-f` flag is filter, not format.

For me this restores thumbnails in the web interface for HEIC images, as I don't have a recent version of imagemagick on my machine.


aviau:
@bradfitz review plz!

aviau:
Turns out this is not true, something else fixed it.

PR #1371, opened 2021-01-22
images/haveImageID: use docker images, not inspect

aviau:
images/haveImageID: use docker images, not inspect
    
    Docker inspect will not exit 0 if no version of the image is available
    on the machine, leading to the method failing. There is no flag to
    adjust this in docker inspect, the only solution would be to parse the
    error message but that would be fragile.
    
    Instead, we opt to use `docker images` which has a very similar output
    with the exception that it will list all tags available for that image
    and that it will not fail on absent images.

This fixed HEIC thumbnails on my machine as I don't have a recent verion of imagemagick.

Initially I submitted [another PR](https://github.com/perkeep/perkeep/pull/1370) trying to adjust the `docker inspect` flags but then I realised that my PR had fixed nothing, it just started working because I had manually pulled the image.

### DIFF
--- a/internal/images/docker.go
+++ b/internal/images/docker.go
@@ -17,6 +17,7 @@ limitations under the License.
 package images // import "perkeep.org/internal/images"
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


aviau:
@bradfitz review plz!

bradfitz (approved):

bradfitz (approved):

PR #1374, opened 2021-01-25
EntityFetcher: require fingerprint

aviau:
EntityFetcher: require fingerprint
    
    EntityFetcher no longer supports key ids. This isn't required anymore as
    its only user (`SignRequest.Sign()`) now passes fingerprints.

### DIFF
--- a/pkg/jsonsign/sign.go
+++ b/pkg/jsonsign/sign.go
@@ -36,7 +36,7 @@ import (
 )
 
 type EntityFetcher interface {
-	FetchEntity(keyID string) (*openpgp.Entity, error)
+	FetchEntity(fingerprint string) (*openpgp.Entity, error)
 }
 
 type FileEntityFetcher struct {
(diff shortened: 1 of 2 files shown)


aviau:
cc @bradfitz for review

aviau:
ping @bradfitz

arturtamborski (approved):

aviau:
@tgulacsi Can you take a look at this one?

tgulacsi:
I've no obligations, just don't understand the context: why is a fingerprint better than the keyID ?
What has changed?

aviau:
> why is a fingerprint better than the keyID ?

Key IDS are like nicknames and can collide, they shouldn't be used for anything serious. They are a bad way to identify keys. See https://evil32.com/ for examples.

> What has changed?

The function used to accept key ids. It no longer does, and I modified the caller to use the full fingerprint since it has it anyways.

tgulacsi (approved):

PR #1375, opened 2021-01-25
discovery: expose full fingerprint

aviau:
discovery: expose full fingerprint
    
    - The discovery endpoint now returns the full key fingerprint.
    
    - Adapt all of the code that used the keyID to use the fingerprint
      instead.
    
    - Begin implementing a test suite for the sign handler's http API.
    
    - Adapt the documentation.


aviau:
cc @bradfitz for review!

PR #1377, opened 2021-01-29
fs: don't cause io errors on missing blobrefs

aviau:
fs: don't cause io errors on missing blobrefs
    
    - Looking up missing blobrefs would cause fs errors. Instead, simply
      return fuse.ENOENT.
    
    - Add tests for:
       * validating that the mountpoint/<blobref> lookups work
    
       * validating that the mountpount/<blobref> lookups don't fail when
         the blob does not exist.


aviau:
cc @bradfitz for review!

bradfitz (approved):

PR #1378, opened 2021-02-01
pkmountTest: ensure tests don't share cache

aviau:
pkmountTest: ensure tests don't share cache
    
    Also introduce PERKEEP_CACHE_DIR so we can deprecate CAMLI_CACHE_DIR
    at some point.

I had a couple of cache issues while working on my previous fs patch. At some point I reverted my entire patch trying to debug why my tests were still failing until I realised that the tests were using my camlicache, causing unexpected behaviors.

This ensures that all mount tests have their own cache.

### DIFF
--- a/pkg/fs/fs_test.go
+++ b/pkg/fs/fs_test.go
@@ -141,15 +141,7 @@ func pkmountTest(t *testing.T, fn func(env *mountEnv)) {
 	defer log.SetOutput(os.Stderr)
 
 	w := test.GetWorld(t)
-	mountPoint, err := ioutil.TempDir("", "fs-test-mount")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer func() {
-		if err := os.RemoveAll(mountPoint); err != nil {
-			t.Fatal(err)
-		}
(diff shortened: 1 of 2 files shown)


pkg/fs/fs_test.go
```diff
@@ -141,7 +141,7 @@ func pkmountTest(t *testing.T, fn func(env *mountEnv)) {
 	defer log.SetOutput(os.Stderr)
 
 	w := test.GetWorld(t)
-	mountPoint, err := ioutil.TempDir("", "fs-test-mount")
+	mountPoint, err := ioutil.TempDir(t.TempDir(), "fs-test-mount")
```
bradfitz, line 144: `mountPoint := t.TempDir()`

Nothing else needed.

aviau: > // Each subsequent call to t.TempDir returns a unique directory;

Ahhh.  Nice.

aviau:
cc @bradfitz  for a second round of review

bradfitz (approved):

PR #1380, opened 2021-02-05
go-mastodon: rollback to 0.0.4

aviau:
go-mastodon: rollback to 0.0.4
    
    The author has removed the 0.0.5 tag (and commits). Revert to 0.0.4 until
    this is explained:
     - https://github.com/mattn/go-mastodon/issues/133
    
    I've also had to changer the cache action's version on macos due to
    permissions issues. I have raised the issue here:
     - https://github.com/actions/cache/issues/527

The tests are broken right now because of this. Actually they are flaky because it looks like some cache servers still remember the 0.0.5 tag.

### DIFF
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -15,7 +15,7 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
 
-    - uses: actions/cache@v2
+    - uses: actions/cache@v2.1.3
       with:
         path: ~/go/pkg/mod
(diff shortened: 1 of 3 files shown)


aviau:
cc @bradfitz for review

aviau:
This was mostly caused by cache issues and Github has since reverted the cache plugin to the bug-free version.

The absence of the 0.5 tag does not prevent go modules to work since we also pin the commit, so  it seems that this PR is no longer needed. Closing.

PR #1381, opened 2021-02-06
devcam hook: fix bad script path

aviau:
devcam hook: fix bad script path
    
    This would create the hook at `.git/hookspre-commit` instead of creating
    it at `.git/hooks/pre-commit`.


aviau:
cc @bradfitz for review!

aviau:
Confident about this, merging.

PR #1386, opened 2021-03-19
Speed S3 query on startup

lologarithm:
Currently I have both a local filesystem blobstore as well as s3 as a backup. When the service starts up it is querying to see if files exist in s3. However I have no files that start with 'packed' and iterating all 400k sha224 blobs takes FOREVER.

S3 does not have 'folders' only prefixes and so we can query all files from the packed "directory" by adding the s3Storage.dirPrefix to our query.

With this change my server now starts up instantly.

### DIFF
--- a/pkg/blobserver/s3/enumerate.go
+++ b/pkg/blobserver/s3/enumerate.go
@@ -48,11 +48,16 @@ func (sto *s3Storage) EnumerateBlobs(ctx context.Context, dest chan<- blob.Sized
 
 	keysGotten := 0
 
-	err := sto.client.ListObjectsV2PagesWithContext(ctx, &s3.ListObjectsV2Input{
+	lo := &s3.ListObjectsV2Input{
 		Bucket:     &sto.bucket,
 		StartAfter: aws.String(sto.dirPrefix + after),
 		MaxKeys:    maxKeys,
-	}, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
+	}
+	if sto.dirPrefix != "" {
+		lo.Prefix = &sto.dirPrefix
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/blobserver/s3/enumerate.go
```diff
@@ -48,11 +48,16 @@ func (sto *s3Storage) EnumerateBlobs(ctx context.Context, dest chan<- blob.Sized
...
 	keysGotten := 0
 
-	err := sto.client.ListObjectsV2PagesWithContext(ctx, &s3.ListObjectsV2Input{
+	lo := &s3.ListObjectsV2Input{
 		Bucket:     &sto.bucket,
 		StartAfter: aws.String(sto.dirPrefix + after),
 		MaxKeys:    maxKeys,
-	}, func(page *s3.ListObjectsV2Output, lastPage bool) bool {
+	}
+	if sto.dirPrefix != "" {
```
aviau, line 56: Setting an empty prefix should work just fine? Its the default value for a string anyways. Do we really need to check before?

lologarithm: The input is a pointer to a string so I assume that there could be differences in implementation between empty and null string. I didn't want to change what behavior we already had (passing in null string) when there is no prefix.

lologarithm: I am not against changing it to always including the pointer.

aviau: Right, but I am pretty sure an empty prefix works. I'd go with that unless it does not work.

lologarithm: done

aviau:
Thanks for the contribution!

PR #1387, opened 2021-03-28
server/perkeepd: Add hints for newcomers about config source

mikefaille:


aviau:
I don't think that more noise is the solution here. `perkeepd -help` mostly answers your question.

The rest is well documented under [doc/server-config](https://perkeep.org/doc/server-config).

mikefaille:
Considering : 

- the path to configure is generally the full path or the relative path from the working directory
- the actual path to configure with `-configfile` refer to the undocumented variable `CAMLI_CONFIG_DIR` as relative prefix for the config file
- effectively, it's possible to check the documentation but this documentation is fragmented in many place. As example, only from the cli -help, a newcomers can't understand where the config is loaded.

I was proposing to get clearer diagnosis for this case where it's not intuitive to get the right path by only consuming the help. The default behaviour isn't  reusing familiar behaviour known of most end user.  

A typical use case with issue :  
By familiarity with most app cli usage guideline, it's rational to think we can craft the configuration from scratch and having the configuration on the same folder of the binary then referring this configuration with the `-configfile`  flag using relative path argument. On my case, I tried to load the config named `server-config.json` (a name proposed by the documentation) relatively to the `perkeepd` binary on the working directory.  Once `perkeepd` started, there is no error showing something was wrong. But, my custom config file wasn't loaded. I understand what happen by crossing the information from : the `-help`, the documentation site and the source code to understand where can I configure the full path of the configuration directory.

mikefaille:
@aviau  As an alternative to reduce logging for configuration debugging purpose, I propose this PR : https://github.com/perkeep/perkeep/pull/1388/files

mikefaille:
Also, to get a more familiar cli usage for most users, it's not possible to change the flag as without notice, we should not change the behaviour of the flag `-configfile` to not break actual configuration.

PR #1388, opened 2021-03-28
server/perkeepd: Add hints on -help for newcomers about config dir

mikefaille:


aviau:
Please keep the text simple.

```
Config file to use, relative to the Perkeep configuration directory root specified by CAMLI_CONFIG_DIR (<PerkeepConfigPath>).
```

server/perkeepd/perkeepd.go
```diff
@@ -109,6 +110,12 @@ var (
 )
 
 func init() {
+	var flagHeaderText = "Usage of %s\nFor the complete documentation, go to : https://perkeep.org/doc/server-config\n\n"
```
aviau, line 113: I don't think this is needed.

mikefaille:
I just applied your suggestion.

server/perkeepd/perkeepd.go
```diff
@@ -96,9 +96,9 @@ var (
 	flagHelp       = flag.Bool("help", false, "show usage")
 	flagLegal      = flag.Bool("legal", false, "show licenses")
 	flagConfigFile = flag.String("configfile", "",
-		"Config file to use, relative to the Perkeep configuration directory root. "+
-			"If blank, the default is used or auto-generated. "+
-			"If it starts with 'http:' or 'https:', it is fetched from the network.")
+		"Config file to use, relative to the Perkeep configuration directory root specified by this environnement variable CAMLI_CONFIG_DIR defled to $HOME/.config/perkeep. "+
+			"If the value of -configfile is blank, the default is used or auto-generated. "+
```
aviau, line 100: Why add words here? It was fine as-is.

server/perkeepd/perkeepd.go
```diff
@@ -96,9 +96,9 @@ var (
...
+			"If the value of -configfile is starts with 'http:' or 'https:', it is fetched from the network.")
```
aviau, line 101: Why add words here? It was fine as-is.

server/perkeepd/perkeepd.go
```diff
@@ -96,9 +96,9 @@ var (
 	flagHelp       = flag.Bool("help", false, "show usage")
 	flagLegal      = flag.Bool("legal", false, "show licenses")
 	flagConfigFile = flag.String("configfile", "",
-		"Config file to use, relative to the Perkeep configuration directory root. "+
-			"If blank, the default is used or auto-generated. "+
-			"If it starts with 'http:' or 'https:', it is fetched from the network.")
+		"Config file to use, relative to the Perkeep configuration directory root specified by this environnement variable CAMLI_CONFIG_DIR defled to $HOME/.config/perkeep. "+
```
aviau, line 99: Please use the PerkeepConfigDir function,

`Config file to use, relative to the Perkeep configuration directory root specified by the CAMLI_CONFIG_DIR environment varaible (currently <PerkeepConfigDir>)`

server/perkeepd/perkeepd.go
mikefaille: For the line 100-101.
I reread and there is apparent conflict of interpretation with the variable `CAMLI_CONFIG_DIR` and the configuration of `configfile`. I could remove it. You will see the difference.

aviau: Provided that you use the wording that I have suggested, it will be fine.

mikefaille:
Nevertheless, I believe you are correct to use a dynamic variable to find the default config path despite my intention because what I took for granted as the default value of the variable only applies to Linux obviously while Mac OSX, and Windows are supported.
I search on the code and also found that Google Cloud Compute is also supported for storing configs. https://github.com/perkeep/perkeep/blob/58991d16f492a740c20d6d4d534ec3e14b2a5f0d/internal/osutil/gce/gce.go#L49

Unfortunately, this is information is hidden in the code. We should go back to that in the doc.

Because the explanations become large to cover the cases, at this point I believe that the ideal is to update the doc in markdown. I thought that we could have provided help more ready to use, that is to say in the help of the cli. But, I resign myself to saying that it will be necessary to explain that in the markdown.

The markdown doc just supports linux to document the path for example. What could allow a more exhaustive doc nearer the enduser workflow is to start a `man page`. Then, we generate the markdown from the man. For help in the cli, we could put an option to display the man's doc (useful for Windows because there is no man) 
It is true that the `.config` file is specific to Linux. Windows also has its own configs. https://github.com/perkeep/perkeep/blob/58991d16f492a740c20d6d4d534ec3e14b2a5f0d/internal/osutil/paths.go#L191

PR #1396, opened 2021-07-29
move to go:embed

### LINKED ISSUE #1384: `goui.js` breaking the dependency scanner on upgrade
Pulled to `commit a49f6851` this morning, rebuilt the server, and ran into this.

```
Mar 01 12:53:40 archlinux perkeepd[2549427]: 2021/03/01 12:53:40 Could not parse deps for goui.js: bufio.Scanner: token too long
```

Seems to be because I have an old version (maybe a year since my last pull?) of `server/perkeepd/ui/zembed_goui.js.go` lying around which doesn't seem to get cleaned up at any point when pulling new versions (or at least deleting that file and rebuilding made everything work again.)

aviau:
- move to go:embed instead of our own custom solution
- simplify buildsystem
- improve build time
- kinda closes #1384, these files are no longer gitignored


tgulacsi (approved):
A logical next cleanup could be replacing wkfs with fs.

aviau:
Thanks for the review.

I am getting pretty confident about this. There shouldn't be any difference in behaviour.

I'll be on the lookout for complaints :)

> A logical next cleanup could be replacing wkfs with fs.

Good idea, I'll give it a shot.

PR #1395, opened 2021-07-26
pkg/blobserver/diskpacked: Implement TODO about StreamBlobs

tgulacsi:
Optimize StreamBlobs for not reading blobs

PR #1400, opened 2021-07-31
pkg/blobserver/diskpacked cleanup

tgulacsi:
First part of https://github.com/perkeep/perkeep/pull/1395

### DIFF
--- a/pkg/blobserver/diskpacked/diskpacked.go
+++ b/pkg/blobserver/diskpacked/diskpacked.go
@@ -125,13 +125,13 @@ func IsDir(dir string) (bool, error) {
 // files.
 func New(dir string) (blobserver.Storage, error) {
 	var maxSize int64
-	var n, atMax int
 	if dh, err := os.Open(dir); err == nil {
+		var nBlobFiles, atMax int
 		if fis, err := dh.Readdir(-1); err == nil {
 			// Detect existing max size from size of files, if obvious, and set maxSize to that
 			for _, fi := range fis {
 				if nm := fi.Name(); strings.HasPrefix(nm, "pack-") && strings.HasSuffix(nm, ".blobs") {
-					n++
+					nBlobFiles++
 					if s := fi.Size(); s > maxSize {
 						maxSize, atMax = fi.Size(), 0
 					} else if s == maxSize {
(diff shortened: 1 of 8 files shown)


pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -125,9 +125,26 @@ func IsDir(dir string) (bool, error) {
 // files.
 func New(dir string) (blobserver.Storage, error) {
 	var maxSize int64
-	if ok, _ := IsDir(dir); ok {
-		// TODO: detect existing max size from size of files, if obvious,
-		// and set maxSize to that?
+	var n, atMax int
```
aviau, line 128: `n` could have a better name. What is it counting? How about numPacks?

Also I think that  the `var n, atMax int` can be inside the `if` statement. It is only used in that case.

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -125,9 +125,26 @@ func IsDir(dir string) (bool, error) {
...
+					} else if s == maxSize {
+						atMax++
+					}
+				}
+			}
+		}
+	}
+	// Believe to the deduced size only if at least 2 files has that maximum size,
+	// and all files (except one) has the same.
+	if !(atMax > 1 && n == atMax+1) {
```
aviau, line 146: This check can be moved inside the `if` statement. We only need to do this if we have looped over all files. So it can be moved directly after the for loop, inside the same statement.

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -125,9 +125,26 @@ func IsDir(dir string) (bool, error) {
...
+					if s := fi.Size(); s > maxSize {
+						maxSize, atMax = fi.Size(), 0
+					} else if s == maxSize {
+						atMax++
+					}
+				}
+			}
+		}
+	}
+	// Believe to the deduced size only if at least 2 files has that maximum size,
```
aviau, line 144: My english is not the best but I think that `to` is unnecessary here :)

aviau:
Minor comments but the rest looks good to me!

If you want to also add a tests that would be nice, but its straightforward enough that I don't mind merging without.

aviau:
yay, thanks!

PR #1401, opened 2021-07-31
pkg/blobserver/diskpacked implement StreamBlobs TODO

tgulacsi:
Optimize walking blobs for not reading them.
Second part of https://github.com/perkeep/perkeep/pull/1395


aviau:
@tgulacsi Can you rebase?

I intend to review this soon, by the way, hoping for this weekend.

Your renewed activity in perkeep is motivating for me, I'd love to keep collaborating on stuff with you :)

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -614,23 +616,31 @@ func (s *storage) StreamBlobs(ctx context.Context, dest chan<- blobserver.BlobAn
...
-		// TODO: remove this allocation per blob. We can make one instead
-		// outside of the loop, guarded by a mutex, and re-use it, only to
-		// lock the mutex and clone it if somebody actually calls ReadFull
-		// on the *blob.Blob. Otherwise callers just scanning all the blobs
-		// to see if they have everything incur lots of garbage if they
-		// don't open any blobs.
-		data := make([]byte, size)
-		if _, err := io.ReadFull(r, data); err != nil {
-			return err
+		var br *blob.Blob
```
aviau, line 619: We usually use the short `br` for blobrefs, but this is a blob.

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -614,23 +616,31 @@ func (s *storage) StreamBlobs(ctx context.Context, dest chan<- blobserver.BlobAn
...
+		{
+			fn := fd.Name()
+			size, offset := size, offset
+			if _, err = io.CopyN(ioutil.Discard, r, int64(size)); err != nil {
```
aviau, line 623: Does this cause us to read the full blob? Can we just not read it?

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -761,3 +771,77 @@ func (m blobMeta) String() string {
...
+	defer fhc.mu.Unlock()
+	fd, err := os.Open(fn)
+	if err != nil {
+		return nil, err
+	}
+	fh = &fileRefCount{File: fd, refCount: 1}
+	fhc.m[fn] = fh
+	return fh, nil
+}
+func (fhc *fileHandleCache) Put(fh *fileRefCount) {
```
aviau, line 810: I may just be unfamiliar with the naming for similar datastructures, but what does "put" mean here?

I see that it decreases the refcount and possibly remove the file handle from the cache.

Looking at the code it seems that the maximum number of file handles isn't really enforced?

What if they all have a refcount of 1?  What is going to stop us from caching an unlimited of file handles if the user does not read a single blob?

PR #1404, opened 2021-08-05
Make staticcheck pass

tgulacsi:
Dead code elimination,
Move to non-deprecated methods,
check unchecked errors/variables,
error string formatting.


aviau:
This looks like its going to be a very large diff. Can we compromise and enable it package-per-package?

tgulacsi:
Aye, aye, captain!

PR #1564, opened 2021-08-10
Lint time since until

tgulacsi:
Use time.Since, time.Until, add staticcheck S1012,S1024 checks.


aviau:
Can we move the staticcheck config to `staticcheck.conf` ?

Ideally on multiple lines so we don't have one long non-git-friendly line.

something like this:
```
checks = [
  "-U1000",
  "-S1012",
  "-S1024",
]
```

aviau:
Merged in 8df8a3657d46a8cff055fed83e88451c544e262c

PR #1571, opened 2021-11-07
Set far-future caching header on blob responses

surma:
(First time contributing, and my Go is rusty!)

Considering the blob store is content-addressed, it seems to make sense to set a far-future caching header by default. The web UI should automatically benefit from this, too.

WDYT?

### DIFF
--- a/pkg/blobserver/gethandler/get.go
+++ b/pkg/blobserver/gethandler/get.go
@@ -33,6 +33,11 @@ import (
 	"go4.org/readerutil"
 )
 
+const (
+	// TODO: Expose this to the config somehow?
+	HTTP_CACHE_DURATION = 10 * 356 * 24 * time.Hour
+)
+
 var getPattern = regexp.MustCompile(`/camli/` + blob.Pattern + `$`)
 
 // Handler is the HTTP handler for serving GET requests of blobs.
@@ -84,6 +89,7 @@ func ServeBlobRef(rw http.ResponseWriter, req *http.Request, blobRef blob.Ref, f
 	}
 	defer rc.Close()
 	rw.Header().Set("Content-Type", "application/octet-stream")
+	rw.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d", int(HTTP_CACHE_DURATION.Seconds())))
 
 	var content io.ReadSeeker = readerutil.NewFakeSeeker(rc, int64(size))
 	rangeHeader := req.Header.Get("Range") != ""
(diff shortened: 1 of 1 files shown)


zenhack:
Hm, maybe we should do `Cache-Control: immutable` as well?

surma:
`Cache-Control: immutable` is [not supported by Chrome](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#browser_compatibility), sadly.

zenhack:
That's unfortunate, but it probably still makes sense to add it for browsers that do support it (you still need to set max-age for it to work, so it's not an either/or anyway).

surma:
I’m not 100% sure what happens if a browser without support for `immutable` receives `Cache-Control: max-age=999, immutable`. Will it ignore just the `immutable` part or will it disregard the entire header? If it’s the former, then it absolutely should be added.

pkg/blobserver/gethandler/get.go
```diff
@@ -84,6 +89,7 @@ func ServeBlobRef(rw http.ResponseWriter, req *http.Request, blobRef blob.Ref, f
 	}
 	defer rc.Close()
 	rw.Header().Set("Content-Type", "application/octet-stream")
+	rw.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d", int(HTTP_CACHE_DURATION.Seconds())))
```
surma, line 92:
```suggestion
	rw.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d, immutable", int(HTTP_CACHE_DURATION.Seconds())))
```

zenhack:
Browsers are supposed to ignore directives they don't understand. I would be surprised if Chrome got that wrong. But I have not tested.

surma:
My colleague says the same thing. Let's add it!

pkg/blobserver/gethandler/get.go
```diff
@@ -33,6 +33,11 @@ import (
 	"go4.org/readerutil"
 )
 
+const (
+	// TODO: Expose this to the config somehow?
```
aviau, line 37: Do we need that todo? Why would one want to configure this?

aviau:
LGTM, one small nit.

Let me know if you want me to merge.

pkg/blobserver/gethandler/get.go
surma:
```suggestion
```

surma:
I’d be stoked to see it merged :)

aviau:
Thank you for your contribution :)

PR #1575, opened 2021-11-23
web-ui/importer: use relative links for importers and accounts

NiklasMerz:
In some network configurations it's annoying that the links in the importer
web ui point to the configured base URL. If you use VPN like tailcale you
might not have the same URL (like home network IP)to access the server.

This changes the links in the web UI to relatives URLs. Since this method
is only used in the web UI it should be pretty safe.

### DIFF
--- a/pkg/importer/importer.go
+++ b/pkg/importer/importer.go
@@ -1163,7 +1163,7 @@ func (ia *importerAcct) AccountObject() *Object { return ia.acct }
 func (ia *importerAcct) RootObject() *Object    { return ia.root }
 
 func (ia *importerAcct) AccountURL() string {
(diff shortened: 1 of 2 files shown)


pkg/importer/importer.go
```diff
@@ -1163,7 +1163,7 @@ func (ia *importerAcct) AccountObject() *Object { return ia.acct }
 func (ia *importerAcct) RootObject() *Object    { return ia.root }
 
 func (ia *importerAcct) AccountURL() string {
-	return ia.im.URL() + "/" + ia.acct.PermanodeRef().String()
```
aviau, line 1166: Could the base url have a prefix here?

PR #1579, opened 2021-12-12
Add a "static site" handler to serve resources from a camliRoot

angelcabo:
This PR adds a static site app handler to enable serving static web resources, like a single page app, from a configured camliRoot node.

Once you configure the handler, you can "install" the app by uploading the files to a camliRoot.

Upload files with: `pk-put file --permanode --title='Static App Source' ./test-app-source`

You can also configure a search root for the app handler to enable searching via the app route:
`pk-put permanode --title='Test App Search Root'`

```
"/static-app/": {
	"handler": "app",
	"enabled": ["_env", "${CAMLI_STATICAPP_ENABLED}"],
	"handlerArgs": {
		"prefix": "/static-app/",
		"serverListen": "localhost:3179",
		"program": "staticapp",
		"appConfig": {
			"staticRoot": "Static App Source",
			"searchRoot": "Test App Search Root"
		}
	}
}
```

Then the app endpoint will support searching with queries limited to nodes in the specified search root:
```bash
$ devcam server -staticapp
$ curl -X POST http://localhost:3179/test-app/search -d' { "expression" : "is:image", "describe": {} } '
```

### DIFF
--- a/app/staticapp/main.go
+++ b/app/staticapp/main.go
@@ -1,0 +1,297 @@
+/*
+Copyright 2021 The Perkeep Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// The Static App application serves static web resources stored in a Perkeep camliRoot directory.
+package main // import "perkeep.org/app/staticapp"
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"perkeep.org/pkg/app"
+	"perkeep.org/pkg/blob"
+	"perkeep.org/pkg/blobserver/localdisk"
+	"perkeep.org/pkg/buildinfo"
+	"perkeep.org/pkg/cacher"
+	"perkeep.org/pkg/schema"
+	"perkeep.org/pkg/search"
+	"perkeep.org/pkg/webserver"
+	"runtime"
+	"strings"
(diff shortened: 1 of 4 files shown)


aviau:
Hey, that looks pretty cool.

Looking at this, I am thinking of the following:
- We should rename the current `fs` package to `fuse`.
- We should create an `fs` package that implements `io.FS` on top of a fetcher (and/or client).
- Your app your use that new package. Removing most of the logic from it.

I'd be willing to help with some of that work if you also agree that it makes sense.

angelcabo:
@aviau I think that's an interesting idea. Do you think the idea would be that this new implementation could eventually replace the pk-mount backend? I know that's built with fuse, but I'm not sure if the implementations would align well. I took a cursory glance at the mount code and it seems like that uses a 3rd party FS interface for emulating the filesystem. Maybe we would just keep them separate, but I could see some overlap between the filesystem emulation currently built in the existing fuse package, and whatever would be added to this proposed fs package. What are your thoughts?

angelcabo:
In other words, I hadn't thought about it until your comment, but maybe it would have made more sense if I had initially used the existing `CamliFileSystem` implementation via `NewRootedCamliFileSystem` to fetch static files, rather than using my own lookup implementation and schema.NewFileReader directly. Had I done that, is there still value in wrapping the existing CamliFileSystem behind the new io/fs abstraction interfaces? Would it make sense to just make the existing CamliFileSystem implement the io/fs interfaces? Or does it still make sense to you to start fresh and let the fuse-specific filesystem implementation sit alongside some new filesystem abstraction layer?

angelcabo:
Actually, on further thought, it probably doesn't make sense to try to make the existing CamliFileSystem conform both to a `io/fs` implementation and a `fuse.FS` implementation given the interfaces overlap in conflicting ways. Rather, a new filesystem abstraction built with the `io/fs` package could possibly be used as an implementation detail by the existing fuse-based CamliFileSystem, but regardless would be separate from the existing fuse package.

aviau:
> ways. Rather, a new filesystem abstraction built with the io/fs package could possibly be used as an implementation detail by the existing fuse-based CamliFileSystem, but regardless would be separate from the existing fuse package.

Yes, exactly.

The current FUSE package should eventually be switch to a user of the FS interface. If we implement io/fs we can get WebDAV almost for free too, and that would help mac users since basil no longer supports macos.

PR #1581, opened 2021-12-24
Use modernc.org/sqlite

tgulacsi:
This way it can be enabled by default, as it is a cgo-free, Go-only package.
No need for build tags, conditional compilation (whether libsqlite3-dev is installed).

My more distant purpose is to have an sqlite-based index instead of the current memory-based,
which consumes 4GiB for me now.


zenhack:
I'm having trouble finding any succinct explanation of what the `modernc.org` packages are all about, but as far as I can tell it looks like they're using https://gitlab.com/cznic/ccgo to translate the source of sqlite to Go and then just committing it to the repo. I have several objections to this:

- I'd like to move *away* from versioning blobs, building things from source properly instead.
- Much more importantly, I do not trust that tool. This seems like it's introducing a ton of complexity, I seriously, seriously doubt that it has seen enough battle testing to be trustworthy, and I would not be comfortable entrusting my data to a version of sqlite that has been run through that.

I think putting up with cgo is a much better option.

tgulacsi:
cgo transpiles Go to C, to be able to compile everything with gcc.
modernc.org/sqlite is a transpilation of the C amalgamation of sqlite3 to Go, so the standard Go compiler can compile it.

I can't decide which is more complex, but you're right that modernc.org/sqlite is younger - though it passes the same tests as sqlite3 does...
Getting rid of cgo may earn a few complexity/risk points.

mohammed90:
> cgo transpiles Go to C, to be able to compile everything with gcc.

I think the compiler uses linker magic to coordinate the Go/C calls rather than transpiling Go to C. It compiles the C bits with gcc, the Go bits with the Go compiler, then the linker does its magic.

zenhack:
Quoting Mohammed Al Sahaf (2021-12-24 12:40:37)
>    I think the compiler uses linker magic to coordinate the Go/C calls
>    rather than transpiling Go to C. It compiles the C bits with gcc, the
>    Go bits with the Go compiler, then the linker does its magic.

This is more or less correct, though I'm not sure I'd agree with the use
of the word "magic;" the C ABI isn't that complicated, and what cgo is
doing is pretty similar to what basically every other language's C FFI
does.

make.go
```diff
@@ -52,7 +53,7 @@ var haveSQLite = checkHaveSQLite()
 
 var (
 	embedResources   = flag.Bool("embed_static", true, "Whether to embed resources needed by the UI such as images, css, and javascript.")
-	sqlFlag          = flag.String("sqlite", "false", "Whether you want SQLite in your build: true, false, or auto.")
+	sqlFlag          = flag.String("sqlite", "false", "DEPRECATED. Whether you want SQLite in your build: true, false, or auto.")
```
bradfitz, line 56: clarify more what this means now. Is this flag ignored, always true, etc?

bradfitz (commented):
Thanks! I've been meaning to do this too. I use modernc.org/sqlite for other things and love it.

make.go
tgulacsi: Always true.
Shall we just remove this, or formulate this differently?

Are the supported [targets](https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures) enough for Perkeep? If no, then we may have to keep this machinery for non-cgo-free-sqlite platforms...

bradfitz: Let's just remove the flag. Those targets look fine.

server/perkeepd/perkeepd.go
```diff
@@ -66,15 +66,16 @@ import (
...
 	// Indexers: (also present themselves as storage targets)
 	// KeyValue implementations:
 	_ "perkeep.org/pkg/sorted/kvfile"
 	_ "perkeep.org/pkg/sorted/leveldb"
 	_ "perkeep.org/pkg/sorted/mongo"
 	_ "perkeep.org/pkg/sorted/mysql"
 	_ "perkeep.org/pkg/sorted/postgres"
-	"perkeep.org/pkg/sorted/sqlite" // for sqlite.CompiledIn()
 
+	// for sqlite.CompiledIn()
```
bradfitz, line 78: delete this line too?

zenhack:
Someone reported on the mailing list that perkeep no longer builds with `GOOS=openbsd` and `GOARCH=amd64`. If @bradfitz has had good luck with this package that eases my initial concerns, but it sounds like this is breaking users; do we want to consider reverting this unless and until `modernc.org/libc` supports openbsd?

PR #1588, opened 2022-02-19
encrypt: use age encryption

MichaHoffmann:
This PR switches the encryption method of the `encrypt` blobserver to [age](https://github.com/FiloSottile/age), since its easier to use and get right. This constitutes a breaking change in this blobserver in the current version for the sake of simplicity. It could be made to be backwards compatible by comparing the version of the blob and choosing an appropriate encryption/decryption method though if that is required!


MichaHoffmann:
Test failure is not related to the change. Seems to be some flakiness in `pk-mount`.

bradfitz (approved):

PR #1590, opened 2022-02-21
fuse: fix truncation of files

### LINKED ISSUE #1218: Incorrect truncation of files in the FUSE filesystem
I get surprising results when I try to write to already existing files in the FUSE filesystem, but the example below sums it up nicely:

```sh
$ # Perkeep FS is mounted at ~/Perkeep
$ mkdir -p ~/Perkeep/roots/test && cd $_
$ printf '%s\n' {1..3} > seq.txt  # let's create a file with numbered lines
$ cat seq.txt
1
2
3
$ # so far, so good
$ echo x > seq.txt  # seq.txt should now contain a single line: x
$ cat seq.txt
x
2
3
$ # wtf?
```

```sh
$ perkeepd --version
perkeepd version: master, 2018-09-14-ca7f2e57e3
Go version: go1.11 (linux/amd64)
```

MichaHoffmann:
This PR fixes a bug in pkg/fs/mut.go where the size attribute was ignored on truncations.

Fixes #1218

### DIFF
--- a/pkg/fs/fs_test.go
+++ b/pkg/fs/fs_test.go
@@ -253,6 +253,52 @@ func TestReadFileFromRoot(t *testing.T) {
 	})
 }
 
+func TestTruncateFile(t *testing.T) {
+	condSkip(t)
+	inEmptyMutDir(t, func(env *mountEnv, rootDir string) {
+		tmpFile, err := ioutil.TempFile(rootDir, "camlitest")
+		if err != nil {
+			t.Fatal(err)
+		}
+		if _, err = tmpFile.WriteString("hello world from test"); err != nil {
+			t.Fatal(err)
+		}
+		if err = tmpFile.Close(); err != nil {
+			t.Fatal(err)
+		}
+
+		const truncateAt = 6
+
+		tmpFile, err = os.OpenFile(tmpFile.Name(), os.O_RDWR, 0644)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err = tmpFile.Truncate(truncateAt); err != nil {
+			t.Fatal(err)
+		}
+		if stat, err := tmpFile.Stat(); err != nil {
+			t.Fatal(err)
+		} else if stat.Size() != truncateAt {
+			t.Fatalf("file size = %d, want %d", stat.Size(), truncateAt)
+		}
+
+		if _, err = tmpFile.WriteAt([]byte("perkeep"), truncateAt); err != nil {
+			t.Fatal(err)
+		}
+		if err = tmpFile.Close(); err != nil {
+			t.Fatal(err)
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/fs/fs_test.go
```diff
@@ -253,6 +253,44 @@ func TestReadFileFromRoot(t *testing.T) {
...
+		}
+
+		tmpFile, err = os.OpenFile(tmpFile.Name(), os.O_RDWR, 0644)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err = tmpFile.Truncate(6); err != nil {
+			t.Fatal(err)
+		}
+		if _, err = tmpFile.WriteAt([]byte("perkeep"), 6); err != nil {
```
bradfitz, line 277: how about a `tmpFile.Stat` here first and make sure the size is 6 first?

at that point there are 3 uses of '6' so maybe `const truncPos = 6` locally

pkg/fs/fs_test.go
```diff
@@ -253,6 +253,44 @@ func TestReadFileFromRoot(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		if err = tmpFile.Close(); err != nil {
+			t.Fatal(err)
+		}
+
+		contents, err := ioutil.ReadFile(tmpFile.Name())
```
bradfitz, line 284: got, err := ...

pkg/fs/fs_test.go
```diff
@@ -253,6 +253,44 @@ func TestReadFileFromRoot(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if cont := string(contents); cont != "hello perkeep" {
+			t.Fatalf("expected file content to be 'hello perkeep', got '%s'", cont)
```
bradfitz, line 289: got %q, not '%s'

but in Go, the "got" usually comes first, and then the want.

    if want := "hello perkeep"; string(got) != want {
            t.Fatalf("file contents = %q; want %q", got, want)
    }

Btw, if you do the `string([]byte)` conversion later, in arguments to `==` or `!=`, it doesn't end up allocating. (Not that it matters in this case)

bradfitz (approved):

pkg/fs/fs_test.go
MichaHoffmann: good point, done.

pkg/fs/fs_test.go
MichaHoffmann: done together with the next comment

pkg/fs/fs_test.go
MichaHoffmann: good information about the allocations, i did not know that, thanks! had to look up '%q' too, i have to admit! I updated everything to your comments in the latest version!

bradfitz (approved):

PR #1592, opened 2022-02-22
cleanup: remove pkg/test/asserts

### LINKED ISSUE #1192: Determine the fate of 'asserts' package
Currently, perkeep.org/pkg/test/asserts contains a warning not to use it in its current state.

I think we have a few options for moving forwards:

1. Replace with an existing other library (I think testify/assert is the most popular)
2. Improve the asserts package; fix whatever issue resulted in its deprecation and begin using them again
3. Don't use any general asserts library, rather inline the checks in tests that use asserts now and create package-specific assertions as needed

Option 1 has the advantage of us maintaining the least code while still being able to remove extra noise from tests. It has the disadvantage of most assert libraries being a pile of reflection which is fragile at times.
Option 2 is only viable if we can actually enumerate the blocking issues with the current asserts package. As far as I can tell, this hasn't been done.
Option 3 may be difficult to automate as easily as 1 or 2, though it's still technically possible. We won't have such helpers available which is a plus or minus depending on taste.

Regardless what we do, I think actually clarifying the issues with the `asserts` package as it exists today is the first step to deciding what to do next.

If the only issue is the `what` messiness, that can now be fixed with `t.Helper`

MichaHoffmann:
This PR removes the deprecated pk/test/asserts.

Suggestion to fix #1192


pkg/test/integration/camget_test.go
```diff
@@ -67,8 +66,9 @@ func TestCamgetSymlink(t *testing.T) {
 	}
 
 	out := test.MustRunCmd(t, w.Cmd("pk-put", "file", srcDir))
-	// TODO(mpl): rm call and delete pkg.
-	asserts.ExpectBool(t, true, out != "", "pk-put")
+	if out == "" {
+		t.Errorf("pk-put: expected output to be non-empty")
```
bradfitz, line 70: Fatalf if you can't proceed (as it appears you can't here).

only use Errorf if the test can usefully continue and maybe report more problems without crashing

Audit the rest too?

bradfitz (commented):
Btw, one of the few non-offensive test helper libraries is github.com/frankban/quicktest which is commonly imported renamed like:

```
import (
      qt "github.com/frankban/quicktest"
)
```

I'm fine if you want to use that too.

pkg/test/integration/camget_test.go
MichaHoffmann: Good catch I'll audit the rest more carefully

MichaHoffmann: Ok i went through the changes and converted most of them to `Fatalf`. There are a few cases where multiple things are tested in the `jsonsign` test where its reasonable to continue on a bad assertions i think

bradfitz (approved):

PR #1594, opened 2022-02-24
cleanup: remove gc package

MichaHoffmann:
This PR removes pkg/gc which is not used currently.


bradfitz:
Well, eventually we do what garbage collection.

Maybe you can wire it up instead? :)

MichaHoffmann:
Ah It is supposed to collect objects that were marked for deletion! I somehow did not get that the first time reading through. Am i correct? Could you give me a pointer where it should go canonically?

MichaHoffmann:
> Well, eventually we do what garbage collection.
> 
> Maybe you can wire it up instead? :)

Ack! Was thinking initially that this package was forgotten. I'll look into wiring it up at some point!

PR #1597, opened 2022-03-12
jsonsign: write armored keyring

aviau:
```
    jsonsign: write armored keyring

    Write Perkeep's keyring as armored for simpler backup to text files.

    The code is backwards-compatible. We try reading the keyring as if it
    was armored and we default to the old format on failure.
```


MichaHoffmann (approved):
Looks good to me

aviau:
> Looks good to me

Thanks!

Issue #1582, opened 2022-01-01
android-app: link to playstore is dead 


MichaHoffmann:
The link to the google play store seems to be dead, is the android app still alive?

bradfitz:
Google regularly changes the Play Store rules and requires certain changes which if you don't do, they remove your app.

I guess they altered the deal.

MichaHoffmann:
Hey @bradfitz,

Thanks for your answer! Is there a plan to get it back into the playstore? Or should it be installed through other means?

PR #1598, opened 2022-03-13
android: return to the android playstore

### LINKED ISSUE #1596: The Perkeep Android app is missing from the Google Play Store
The download link on the website is broken. The app is missing from the store. The alternate version linked on the website appears to be an older version (even uses the older Camlistore name and branding).

MichaHoffmann:
Fixes #1596 #1582 

This PR intends to make the app viable for the android playstore again

* Intents need to get the flag PendingIntent.FLAG_IMMUTABLE/MUTABLE
* Assets are no longer executable, get around that by using
  'extractNativeLibs=true' and executing from the lib path
* Apply several refactoring suggestions by Android Studio
* Rework resources Makefile
* Sideload `pk-put` as a native lib [very hacky]
* rename Camlistore to Perkeep where appropriate


MichaHoffmann:
Should we also use the opportunity to rename it to perkeep in this PR?

MichaHoffmann:
> Should we also use the opportunity to rename it to perkeep in this PR?

ok i just went ahead and renamed it; PR can be reverted if its problematic.

MichaHoffmann:
FileWatcher seems to be broken, at least i dont register updates when a photo is taken, ill debug that soon

mhoffm-aiven:
Automatic Fileupload now works again ( on my machine ). There was a race condition between inotify and android being able to find the file; a few retries fixed that.

bradfitz (approved):

MichaHoffmann:
Closed in favor of #1611

PR #1604, opened 2022-04-28
Fix various typos

luzpaz:
Found via `codespell -q 3 -S ./clients/web/embed,./clients/chrome -L ba,everytime,impres,keypair,msdos,pres,ro,te,ue`


MichaHoffmann:
lgtm, could probably be added to github actions so that its not just a one off

PR #1608, opened 2022-05-04
geocode: add support for openstreetmap geocoding api

### LINKED ISSUE #1342: Support a free geocoding service
Hi,

I see that Google geocoding support is available, however this is a paid service and I only have a minimal amount of coordinates to resolve.

Would it be possible to support other geocoding services that don't charge for 0-X.000 queries? For example[Yandex.Maps](https://yandex.com/dev/maps/geocoder/)? If it's already possible, how to configure it?

Thanks!

willnorris:
Use the OpenStreetMap [Nominatim API](https://wiki.openstreetmap.org/wiki/Nominatim) when no Google Geocoding API key is present.  Nominatim does not require any keys, so we can safely enable it by default, though their terms request no more than 1 qps, which should probably be fine?

Docs and log messages are updated to reflect this behavior.  OSM Attribution is added to the help page, which is not the best but probably okay for now.  In a future change, I'd like to change the About alert to a static page, and move licensing information there.

Fixes #1342

### DIFF
--- a/doc/geocoding.md
+++ b/doc/geocoding.md
@@ -3,12 +3,12 @@
 Geocoding is the process of converting a location name (like `nyc` or
 `Argentina`) into GPS coordinates and bounding box(es).
 
-Perkeep's location search currently requires Google's Geocoding API,
-which now requires an API key. We do not currently provide a default,
-shared API key to use by default. (We might in the future.)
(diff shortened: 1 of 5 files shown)


MichaHoffmann (approved):
lgtm, though we probably cannot guarantee the 1 query per second

PR #1612, opened 2022-06-13
pkg/sign: use local readKeyRing function instead of opengpg's, to handle armored key rings.

### LINKED ISSUE #1601: openpgp: invalid data: tag byte does not have MSB set
2022/04/03 11:15:42 bad request: jsonsign: openpgp.ReadKeyRing of "/Users/hsh/.config/perkeep/identity-secring.gpg": openpgp: invalid data: tag byte does not have MSB set

featherbutt:
Fixes #1601


aviau:
I approved CI. I'll merge if tests pass.

PR #1615, opened 2022-06-20
fuse: misc fixes

MichaHoffmann:
This should fix some of the flakiness in the `perkeep.org/pkg/fs` tests.

* propagate more context cancelations from pkg/client
* convert ErrCancel to fuse.EINTR to enable clients to handle interrupts
* improve interrupt handling in fs_test.go
* fix race in mutdir between rename and first populate

### DIFF
--- a/pkg/fs/util.go
+++ b/pkg/fs/util.go
@@ -1,3 +1,6 @@
+//go:build linux || darwin
+// +build linux darwin
+
 /*
 Copyright 2013 The Perkeep Authors
 
@@ -17,10 +20,13 @@ limitations under the License.
 package fs
 
 import (
+	"context"
 	"errors"
 	"os/exec"
 	"runtime"
 	"time"
+
+	"bazil.org/fuse"
 )
 
 // Unmount attempts to unmount the provided FUSE mount point, forcibly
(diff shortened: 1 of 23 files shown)


pkg/fs/util.go
```diff
@@ -51,3 +56,22 @@ func Unmount(point string) error {
 		return err
 	}
 }
+
+func interruptOrEio(err error) error {
+	if err == nil {
+		return err
```
bradfitz, line 63: return nil

pkg/fs/util.go
```diff
@@ -51,3 +56,22 @@ func Unmount(point string) error {
 		return err
 	}
 }
+
+func interruptOrEio(err error) error {
```
bradfitz, line 61: orEIO

pkg/fs/util.go
MichaHoffmann: :heavy_check_mark:

pkg/fs/util.go
MichaHoffmann: I called it `handleEIOorEINTR` now

MichaHoffmann:
It might still happen that `f.Close()` gets interrupted; but test stability has greatly improved for me when testing it out since most interrupted syscalls seem to be retried in the stdlib.

PR #1619, opened 2022-07-10
deps: bump aws sdk

MichaHoffmann:


bradfitz:
Sure, but any particular reason?

mhoffm-aiven:
> Sure, but any particular reason?

I'm hitting "KeyNotFound" when restoring encryption metadata although the key is there ( might be eventual consistency issues ) when restarting perkeep. Bumping AWS SDK was just first shot in the dark to try solving that issue. I'm also thinking about retrying this error code a few times even though it feels weird.

PR #1620, opened 2022-07-14
android: fix upload thread restarts

MichaHoffmann:
Before this change there is a race between queuing files and uploading them (very pronounced when pressing `Upload All` with most files in the havecache). When starting to enqueue files, if pk-put drains the queue faster we end up restarting the upload thread very frequently leading to `input/output error: file busy` when trying to create/open the havecache during pk-put start. I have not run into this error again with this change.


bradfitz:
(Closing, per video chat; going to do this differently)

Issue #1605, opened 2022-04-29
Add storage example for Google Cloud Storage


malisetti:
Please add an example for GCS at https://perkeep.org/doc/storage-examples
It is very confusing to get the required configuration information from gcs.

kalidor:
Did you look at https://perkeep.org/doc/server-config ? There is a line about GCS which seems pretty explicit:

`googlecloudstorage: "clientId:clientSecret:refreshToken:bucketName[/optional/dir]"`

malisetti:
https://perkeep.org/doc/storage-examples describes how to get configuration information required from AWS and backblaze. I am only asking the same for GCS

PR #1621, opened 2022-07-15
Issue#1605 - Google Cloud Storage storage docs

mccurdyc:
Addresses https://github.com/perkeep/perkeep/issues/1605

### DIFF
--- a/doc/server-config.md
+++ b/doc/server-config.md
@@ -95,7 +95,7 @@ At least one of these must be set:
 * `s3`: "`key:secret:bucket[/optional/dir]`" or
   "`key:secret:bucket[/optional/dir]:hostname`" (with colons, but no quotes).
 * `b2`: "`account_id:application_key:bucket[/optional/dir]:endpoint`"
-* `googlecloudstorage`: "`clientId:clientSecret:refreshToken:bucketName[/optional/dir]`"
+* `googlecloudstorage`: "`clientId:clientSecret:refreshToken:bucketName/optional/dir`"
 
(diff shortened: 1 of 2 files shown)


doc/server-config.md
```diff
@@ -95,7 +95,7 @@ At least one of these must be set:
 * `s3`: "`key:secret:bucket[/optional/dir]`" or
   "`key:secret:bucket[/optional/dir]:hostname`" (with colons, but no quotes).
 * `b2`: "`account_id:application_key:bucket[/optional/dir]:endpoint`"
-* `googlecloudstorage`: "`clientId:clientSecret:refreshToken:bucketName[/optional/dir]`"
+* `googlecloudstorage`: "`clientId:clientSecret:refreshToken:bucketName/optional/dir`"
```
mhoffm-aiven, line 98: From looking at https://github.com/perkeep/perkeep/blob/c501f90cd011f0d0ca73b0193e343f78eba5130a/pkg/blobserver/google/cloudstorage/storage.go#L104 it seems that this directory is truly optional, so we should probably keep the square brackets.

mccurdyc: Sounds good!! Honestly, I dropped the square brackets because they confused me at first. I thought they were used to parse the path. It wasn't until I saw the path that got created with square brackets that I realized it was just for notation.

So if we do plan to keep them, I can note that so that folks don't get confused like me 😄

mhoffm-aiven: Yeah they are only for notation! I think adding a short paragraph about that notation is a good thing.

PR #1623, opened 2022-07-18
index: add test for out of order indexing when verifying permanode signatures

MichaHoffmann:


pkg/index/index_test.go
```diff
@@ -529,6 +529,53 @@ func TestIndexingClaimMissingPubkey(t *testing.T) {
 	}
 }
 
+func TestIndexingPermanodeMissingPubkey(t *testing.T) {
```
bradfitz, line 532: I'm not great at this myself but it's not super obvious, try to leave a comment above the test describing a high level summary of what this test is testing so the user doesn't need to read all the code to try to figure it out or dig through the git history to figure out issue references.

bradfitz (approved):

PR #1624, opened 2022-08-28
encrypt: dont upload and enumerate concurrently

MichaHoffmann:


MichaHoffmann:
nvm, still crashes with:
```
Sep 02 15:35:00 ubuntu perkeepd[24224]: 2022/09/02 15:35:00 Caught panic installer handlers: error instantiating storage for prefix "/b2-enc/", type "encrypt": error scanning metadata on start-up: error with meta blob sha224-3ff458dbc01f8f37d0d486a34536c2ae249b68a8c78b1866959e6741: fetch failed: unexpected EOF
```

PR #1631, opened 2022-12-19
misc: remove sha1 from tests, remove testhooks, minor cleanups

MichaHoffmann:
* removed sha1 hashes from tests
* removed internal/testhooks
* use `runtime.NumCPU` as default reindex maxprocs
* removed code that disables ooo indexing from tests ( since its not used anymore )


pkg/index/index.go
```diff
@@ -422,26 +423,7 @@ func (x *Index) isEmpty() bool {
...
-	reindexMaxProcs.v = n
-}
-
-// ReindexMaxProcs returns the maximum number of concurrent goroutines that are
-// used during reindexing.
-func ReindexMaxProcs() int {
-	reindexMaxProcs.RLock()
-	defer reindexMaxProcs.RUnlock()
-	return reindexMaxProcs.v
-}
```
bradfitz, line 444: This seems unrelated?

bradfitz: Keep the setter/getter and just change the default of 4 to NumCPU?

MichaHoffmann: :heavy_check_mark:

bradfitz (approved):

Issue #1392, opened 2021-04-21
pk-mount unsupported on MacOS


tbrtje:
Hi there,
Apparently there is no support for MacFUSE 4 and later. If it is installed I get the error message 
```FUSE not available```. This seems to be due to bazil.org/fuse lack of support for it.
Unfortunately this isn't just an issue for M1 based systems, as described here #1327 , but an issue for all up to date MacOS Systems.

Unfortunately I have no clue how to implement a fix, but I hope someone with the knowledge and spare time can :)
Best Regards
Thies

tgulacsi:
We're updating bazil.org/fuse to the latest that has support for MacOS:

https://github.com/perkeep/perkeep/pull/1394/commits/482a7f081267e668f1987908b4339fb392368e1f

Unfortunately that does nothing with the OS that has desupported MacFUSE.

https://github.com/rclone/rclone/issues/4393 uses a C-based alternative, 

https://github.com/restic/restic/issues/485 uses WebDAV.

I do have worked on WebDAV for perkeep  some time ago, may be appropriate to resurrect that

tbrtje:
> Unfortunately that does nothing with the OS that has desupported MacFUSE.
MacFUSE is available for Big Sur and Monterey, but the license was changed with version 4.0
Other tools do work with the current version of MacFUSE, so what do you mean exactly?

tgulacsi:
See https://github.com/bazil/fuse/issues/224 and https://colatkinson.site/macos/fuse/2019/09/29/osxfuse/
OSXFUSE is no longer open source, you have to pay for it.
If you will to do so, then pk-mount may work for you.
If not, we have to find another way. That's why I suggested WebDAV: as far as I see, it is supported out of the box
with Windows and MacOS.
Can you confirm that WebDAV would suffice you on MacOS?

GwynethLlewelyn:
These days, it should be possible to use @anacrolix's own fork of bazil/fuse, which is a drop-in replacement (no code changes necessary) which provides the exact same functionality — plus working macFUSE 4+ support.

You can get it from here: https://github.com/anacrolix/fuse

bradfitz:
@GwynethLlewelyn, thanks for the pointer. (I also forgot about https://www.fuse-t.org/)

I recently removed all our macOS-based FUSE support in afb071614f8b2d554af9cff7e597b103860b5078 but it should be easy to restore it and switch to https://github.com/anacrolix/fuse.

anacrolix:
[...]

Issue #1017, opened 2018-01-06
implement chrome os filesystem provider


willnorris:
This would be the ChromeOS equivalent of Perkeep's current fuse support, allowing a Perkeep-backed filesystem to be accessed in ChromeOS's native Files app.

API Docs: https://developer.chrome.com/apps/fileSystemProvider

Example providers:
 - [Google Cloud Storage](https://github.com/beaufortfrancois/cloud-storage-chrome-app)
 - [SMB](https://github.com/GoogleChromeLabs/chromeos_network_file_share)
 - [WebDAV and Amazon S3](https://github.com/google/chromeos-filesystems)
 - [SFTP](https://github.com/yoichiro/chromeos-filesystem-sftp)
 - [Dropbox](https://github.com/yoichiro/chromeos-filesystem-dropbox)

Docs for using GopherJS in Chrome extensions, which is likely the best bet here: https://github.com/fabioberger/chrome

bradfitz:
@tgulacsi, this is another reason that we should use the generic "fslib" approach you'd proposed. But I don't like "lib" in the name. Maybe we call the generic package "pkfs" instead.

bradfitz:
(Then pk-webdav, pk-mount for FUSE, and this via GopherJS could share the same pkfs implementation.)

Issue #38, opened 2011-07-10
webdav server


bradfitz:
<pre>For OS X fs mounting.  Use same interfaces as FUSE, probably.</pre>

saljam:
golang.org/x/net/webdav is shaping up nicely (still missing some bits, like PROPPATCH, but it's getting there) so this might be a good time to revive this.

Do we want this as a Camlistore app, or baked in Camlistore proper? The only benefit of the latter option I see is piggybacking on Camlistore's authentication.

mpl:
Can't we reuse (and extend if needed) what's in pkg/app/app.go to do the auth for webdav too?

saljam:
Yes, I think that would be the best approach.

mpl:
maybe we could add a Camlistore plugin to https://github.com/mattn/davfs :-)

josephholsten:
cf #146

PR #1633, opened 2022-12-28
app: add initial webdav support

MichaHoffmann:
This PR adds WebDAV support for perkeep as an app. 

caveats:
* only supports static elements (directory and file blobs) pointed at from a camliRoot with "camliPath:x" attributes
*  this  also means that the content is read only
*  master query is too coarse right nowuntil we support the "progeny" relation constraint
* no integration tests yet

I intend on addressing those issues in subsequent PRs by adding support for dynamic directories (permanode with "camliPath:*" attribute or camliContent that points to a directory) and files (permanode with camliContent that points to a file) that will be writeable too. 

I tested this with `cadaver` locally for now.

@tgulacsi you suggested this in the past, so tagging you here. wdyt?

addresses #38 #1392 #1017


MichaHoffmann:
Tests fail because i used go1.17 api in places. Fixing in a bit.

app/webdav/webdav.go
```diff
@@ -0,0 +1,407 @@
...
+				continue
+			}
+			n.sub[fi.FileName] = &webdavFsNode{br: m.BlobRef}
+		}
+	}
+	n.lastRefreshed = time.Now()
+	return nil
+}
+
+type webdavFsFile struct {
```
bradfitz, line 276: maybe get "RO" (for Read Only) into this type name? And then some docs on the type.

app/webdav/webdav.go
```diff
@@ -0,0 +1,407 @@
...
+	sub map[string]*webdavFsNode
+
+	// cache invalidation data
+	static        bool
+	lastRefreshed time.Time
+}
+
+var refreshInterval = 1 * time.Minute
+
+func needsRefresh(n *webdavFsNode) bool {
```
bradfitz, line 159: make this a method on n instead?

and if it requires its n.mu be locked, name it `needsRefreshLocked`?

app/webdav/webdav.go
```diff
@@ -0,0 +1,407 @@
...
+		t = fi.ModTime.Time()
+	}
+	return t
+}
+
+func (f *webdavFsFile) Stat() (os.FileInfo, error) {
+	return f.n.fi, nil
+}
+
+type webdavFsFileInfo struct {
```
bradfitz, line 376: all these "webdav" type prefixes are redundant with the package name. Could just call this "fsFileInfo" ?

MichaHoffmann: :heavy_check_mark:

app/webdav/webdav.go
MichaHoffmann: :heavy_check_mark:

app/webdav/webdav.go
MichaHoffmann: :heavy_check_mark:

Serpentiel (approved):
lgtm

PR #1634, opened 2022-12-28
Delete clients/python directory

ginabythebay:
depends on "ancient and broken Camlistore Python libary" which was removed in 6446c307a1cb5fbf455ea60b3792c6bc4cc2c09d


MichaHoffmann:
lgtm

PR #1636, opened 2022-12-29
ui: remove obsolete code

MichaHoffmann:
`http.ServeContent` figures out the content type on its own, i also checked the headers in my browser with this patch (build with go1.19 though) . 

The comment changes comes from `go fmt`.

### DIFF
--- a/pkg/server/ui.go
+++ b/pkg/server/ui.go
@@ -345,11 +345,11 @@ func (ui *UIHandler) makeClosureHandler(root string) (http.Handler, error) {
 
 // makeClosureHandler returns a handler to serve Closure files.
 // root is either:
-// 1) empty: use the Closure files compiled in to the binary (if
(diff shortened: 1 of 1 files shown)


pkg/server/ui.go
```diff
@@ -515,11 +515,6 @@ func ServeStaticFile(rw http.ResponseWriter, req *http.Request, root fs.FS, file
 	if fi, err := f.Stat(); err == nil {
 		modTime = fi.ModTime()
 	}
-	// TODO(wathiede): should pkg/magic be leveraged here somehow?  It has a
-	// slightly different purpose.
-	if strings.HasSuffix(file, ".svg") {
-		rw.Header().Set("Content-Type", "image/svg+xml")
-	}
```
bradfitz, line 522: Merge this one after Go 1.19 is our minimum.

PR #1638, opened 2022-12-30
camlistore.org -> perkeep.org

ginabythebay:
fix up some links

There are still some references to camlistore.org that I don't understand so left alone.


bradfitz:
Thanks!

Issue #1037, opened 2018-01-13
Better documentation to advanced use case for the search engine


gdeverlant:
Greetings again!

While trying to get to know Perkeep and the search engine I could not do much because of the poor documentation. I adapted Perkeep to work inside of a non-linear video editing browser and I had to bother lots of people trying to figure out how the search is working.

I'm aware that the search capabilities of Perkeep is very powerful but no one can use its full potential.

Please add a better documentation with complex use cases.

Thank you for Perkeep!

adamdecaf:
There's a page for this now (https://perkeep.org/doc/search-ui), but I assume it's incomplete. Is there a good place to start? I've been trying searches like: 

https://github.com/perkeep/perkeep/search?utf8=%E2%9C%93&q=%22search.SearchQuery%22&type=

PR #1640, opened 2022-12-31
demonstrate query using curl

ginabythebay:
It took me a while to figure this out, perhaps it will help other people to have an example

### DIFF
--- a/clients/curl/example.sh
+++ b/clients/curl/example.sh
@@ -4,6 +4,7 @@
 
 # Configuration variables here:
 BSHOST=localhost:3179/bs
+SEARCHHOST=localhost:3179/my-search
 BSUSER=user
 BSPASS=foo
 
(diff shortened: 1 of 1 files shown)


clients/curl/example.sh
```diff
@@ -40,3 +40,6 @@ curl -v -u $AUTH http://$BSHOST/camli/enumerate-blobs?limit=1
 
 # List offset -- 200 with list of no blobs
 curl -v -u $AUTH http://$BSHOST/camli/enumerate-blobs?after=sha1-126249fd8c18cbb5312a5705746a2af87fba9538
+
+# Search for all images
+echo '{"expression":"is:image","describe":{"rules":[{"attrs":["camliContent"]}]}}' | curl -v  --data-binary  @- -u $AUTH http://localhost:3179/my-search/camli/search/query
```
MichaHoffmann, line 45: Since the other examples use `$BSHOST`; maybe we should use something like `$SEARCHHOST`

MichaHoffmann: Also, just in case you didnt know, you could use the perkeep cli for this too `pk search "is:image"`

ginabythebay: > Also, just in case you didnt know, you could use the perkeep cli for this too `pk search "is:image"`

Yep and I think that is useful too!

I'm just trying to record and share how to do it over http.  I couldn't find anything about search here https://perkeep.org/doc/protocol/.   I'm not prepared to fill that gap at this point, but I would like to move towards that.  FWIW, issue #1037 is asking for that as well.

MichaHoffmann: Yeah i think its a useful change, what do you think about using `$SEARCHHOST` similar to `$BSHOST` in the other examples?

ginabythebay: Added

MichaHoffmann (approved):
lgtm

PR #1644, opened 2023-01-08
index: add receive hook

MichaHoffmann:
This removes the need for the `cond` + `replica` storage configuration and hooks up the index and its backing storage using the blobhub. Also fixed a bug where serverinit installs a wrapper instead of the storage which causes the receive hooks to be fired for the wrapper but not for the storage itself causing us to mix sync and index for it.


MichaHoffmann:
Windows failures seem genuine, but i dont understand them yet.

pkg/blobserver/handlers/remove.go
```diff
@@ -28,9 +28,9 @@ import (
 
 const maxRemovesPerRequest = 1000
 
-func CreateRemoveHandler(storage blobserver.Storage) http.Handler {
+func CreateRemoveHandler(storage blobserver.Storage, config *blobserver.Config) http.Handler {
```
bradfitz, line 31: add docs to these methods? how's this config used? send separately?

pkg/blobserver/handlers/stat.go
```diff
@@ -29,22 +29,17 @@ import (
 	"perkeep.org/pkg/blobserver/protocol"
 )
 
-func CreateStatHandler(storage blobserver.BlobStatter) http.Handler {
+func CreateStatHandler(storage blobserver.BlobStatter, config *blobserver.Config) http.Handler {
```
bradfitz, line 32: same

pkg/blobserver/blobhub.go
```diff
@@ -35,13 +36,13 @@ type BlobHub interface {
 	//
 	// If any synchronous receive hooks are registered, they're run before
 	// NotifyBlobReceived returns and their error is returned.
-	NotifyBlobReceived(blob.SizedRef) error
+	NotifyBlobReceived(context.Context, blob.SizedRef) error
```
bradfitz, line 39: send all this context plumbing in its own PR(s)?

pkg/blobserver/blobhub.go
```diff
@@ -35,13 +36,13 @@ type BlobHub interface {
...
 
 	// AddReceiveHook adds a hook that is synchronously run
 	// whenever blobs are received.  All registered hooks are run
 	// on each blob upload but if more than one returns an error,
 	// NotifyBlobReceived will only return one of the errors.
-	AddReceiveHook(func(blob.SizedRef) error)
+	AddReceiveHook(func(context.Context, blob.SizedRef) error)
```
bradfitz, line 45: and some docs on how the context is used?

pkg/blobserver/files/osfs_windows.go
```diff
@@ -17,21 +17,33 @@ limitations under the License.
...
+		syscall.ERROR_ACCESS_DENIED,
+	} {
+		if errors.Is(linkErr, errno) {
+			found = true
+			break
+		}
+	}
+	if !found {
 		return err
 	}
```
bradfitz, line 48: send separately with a test? document/justify the access denied on rename thing.

PR #1645, opened 2023-01-10
schema: address todo about non-whitespace characters after schema blob

MichaHoffmann:


bradfitz:
Sent #1646 as an alternative, fixing parseSuperset to not allow trailing junk instead.

bradfitz:
We went with #1646 instead.

PR #1660, opened 2023-12-30
go.mod: remove old replaces, update deps, use Go 1.21

bradfitz:
Signed-off-by: Brad Fitzpatrick <brad@danga.com>


MichaHoffmann (approved):
lgtm

PR #1661, opened 2023-12-30
pkg/deploy/gce: remove GCE deploy support

bradfitz:
We used to have a web-based Perkeep launcher at perkeep.org/launch
that created a GCE-based Perkeep instance for users, where they pay
Google for compute time. (One of those "one click deploy" template things)

Unfortunately, Google broke their APIs for doing the third party VM
creations and we disabled it some years ago. But the code remains. And
now, updating it again, we find that they've broken it again:

   Error: pkg/deploy/gce/deploy.go:358:4: servicemanagement.NewServicesService(s).Enable undefined (type *servicemanagement.ServicesService has no field or method Enable)

It's not worth fighting Google's API breakages. Just remove the GCE
launcher support as it's been unused for years.

We can always resurrect this code from git if really needed. But a
Digital Ocean or Fly launcher would probably be much easier.


bradfitz:
cc @mpl

PR #1667, opened 2024-01-01
internal, blobserver: add robutio package and use in osfs

MichaHoffmann:
Windows renames can be flaky, which can lead to problems when we tried to add a receive hook for the indexer. To address this we copy the robustio package from
https://github.com/golang/go/tree/master/src/cmd/go/internal/robustio.


internal/robustio/robustio.go
```diff
@@ -0,0 +1,11 @@
+package robustio
```
bradfitz, line 1: Add a package doc here, per https://go.dev/wiki/CodeReviewComments#package-comments

```
// Package robustio ....
package robustio
```

bradfitz (approved):

bradfitz:
> add robutio package

Typo in PR title

Issue #1668, opened 2024-01-02
Add Tailscale support


bradfitz:
Tracking bug to add @tailscale support.

We can use Tailscale for all of:

* getting a LetsEncrypt cert (including a nice hostname),
* for authentication
* for authorization (via grants: https://tailscale.com/blog/acl-grants / https://github.com/tailscale/tailscale/issues/4217)
* sharing, via node sharing, somehow

/cc @willnorris @mpl @MichaHoffmann

quinn:
Could tailscale be used for peer to peer sharing? I was looking at this: https://github.com/meejah/fowl for trying to create p2p shares

PR #1670, opened 2024-01-02
Remove CamliNetIP, camnetdns, gpgchallenge

bradfitz:
These were all part of an earlier effort to let it be easy for people
to easily deploy their own Perkeep servers and get them on the
internet. The idea was that we'd run a small DNS server which would
map from GPG public keys to their rented cloud IPs which users would
prove to us with the gpgchallenge stuff.

The recently added Tailscale support (see 43f34e5cc511, #1668) makes
most of this redundant.

Also I'd stopped running this infrastructure ages ago and removed the
launcher code recently in b5823a65b9587 (and disabled it in
c9f78d02adf). So this was all basically dead code.


bradfitz:
cc @mpl

PR #1676, opened 2024-01-14
all: move from deprecated x/crypto/openpgp to ProtonMail fork

bradfitz:
See https://github.com/golang/go/issues/44226 for background.

The ProtonMail fork is the same API (and code mostly).

Differences are documented at https://github.com/ProtonMail/go-crypto/issues/21

Signed-off-by: Brad Fitzpatrick <brad@danga.com>


bradfitz:
Looks like this is wrong:

```go
func writeKeyProperties(selfSignature *packet.Signature, creationTime time.Time, keyLifetimeSecs uint32, config *packet.Config) error {
	advertiseAead := config.AEAD() != nil

	selfSignature.CreationTime = creationTime
	selfSignature.KeyLifetimeSecs = &keyLifetimeSecs
```

That's setting a pointer to 0 instead of leaving the optional field as nil.

PR #1693, opened 2025-04-14
Diskpacked max open files

tgulacsi:
Limit max open files in diskpacked blobserver.

It has opened (and kept open) all pack-%05d.blobs file.

With this change, an LRU cache is used with 80% of MaxFDs (or of 1024).

### DIFF
--- a/pkg/blobserver/diskpacked/diskpacked.go
+++ b/pkg/blobserver/diskpacked/diskpacked.go
@@ -43,9 +43,12 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"sync"
 
+	"perkeep.org/internal/lru"
+	"perkeep.org/internal/osutil"
 	"perkeep.org/pkg/blob"
 	"perkeep.org/pkg/blobserver"
 	"perkeep.org/pkg/blobserver/local"
@@ -79,19 +82,22 @@ func (d debugT) Println(args ...interface{}) {
 const defaultMaxFileSize = 512 << 20 // 512MB
 
 type storage struct {
-	root        string
-	index       sorted.KeyValue
-	maxFileSize int64
+	index sorted.KeyValue
 
 	writeLock io.Closer // Provided by lock.Lock, and guards other processes from accessing the file open for writes.
 
 	*local.Generationer
 
+	writer      *os.File
+	fdCache     *lru.Cache
+	root        string
+	maxFileSize int64
+
+	fileCount int
+	size      int64
+
 	mu     sync.Mutex // Guards all I/O state.
 	closed bool
-	writer *os.File
-	fds    []*os.File
-	size   int64
 }
 
 func (s *storage) String() string {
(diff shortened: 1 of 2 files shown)


tgulacsi:
[tests/macos / test (1.21.x, macos-latest) (pull_request)](https://github.com/perkeep/perkeep/actions/runs/14449983367/job/40519990097?pr=1693)Failing after 3m

fd count mismatch.

I do have a runtime.GC() after Closeing the cached fds, but the finalizers may or may not run.
Maybe runtime.AddCleanup would be less flaky, but that needs go1.24.
Shall we
a) skip this test on MacOS,
b) eiiminate this test (and runtime.GC()),
c) or add some waits (with channels or WaitGroup)

?

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -214,30 +241,39 @@ func newFromConfig(_ blobserver.Loader, config jsonconfig.Obj) (storage blobserv
...
-func (s *storage) openForRead(n int) error {
-	if n > len(s.fds) {
-		panic(fmt.Sprintf("openForRead called out of order got %d, expected %d", n, len(s.fds)))
+func (s *storage) openForRead(n int) (*os.File, error) {
+	if n > s.fileCount+1 {
+		panic(fmt.Sprintf("openForRead called out of order got %d, expected %d", n, s.fileCount))
 	}
 
 	fn := s.filename(n)
+	if f, ok := s.fdCache.Get(fn); ok {
```
bradfitz: why not make this fdCache keyed by the integer value instead of by the string you have to allocate all the time right now? (that filename method allocates)

pkg/blobserver/diskpacked/diskpacked.go
```diff
@@ -79,19 +81,22 @@ func (d debugT) Println(args ...interface{}) {
...
 
 	*local.Generationer
 
+	writer      *os.File
+	fdCache     *sieve.Sieve[string, *os.File]
+	root        string
+	maxFileSize int64
+
+	fileCount int
+	size      int64
```
bradfitz, line 97: are these no longer guarded by `mu`? If they still are, they should be under the `mu sync.Mutex` line, in that block.

internal/sieve/sieve.go
```diff
@@ -0,0 +1,278 @@
+// sieve.go - SIEVE - a simple and efficient cache
+//
+// (c) 2024 Sudhi Herle <sudhi@herle.net>
+//
+// Copyright 2024- Sudhi Herle <sw-at-herle-dot-net>
+// License: BSD-2-Clause
+//
+// If you need a commercial license for this work, please contact
+// the author.
```
bradfitz: what does this mean? If it's BSD-2-Clause we don't actually need a license.

internal/sieve/sieve.go
```diff
@@ -0,0 +1,278 @@
+// sieve.go - SIEVE - a simple and efficient cache
```
bradfitz, line 1: This doesn't exist elsewhere we could import & use?

e.g. https://pkg.go.dev/github.com/opencoff/go-sieve

Or because you editing it?

pkg/blobserver/diskpacked/diskpacked.go
tgulacsi: Indeed they're, just this way the struct pads less (betteralign).
I'll reorder it.

internal/sieve/sieve.go
tgulacsi: Copy-paste. Deleted.

internal/sieve/sieve.go
tgulacsi: We can import & use that, just needed a callback on remove (to explicitly close the file, don't use a Finalizer.
For this I needed to edit it, so removed the unneeded mutex and methods (Dump, for example).

pkg/blobserver/diskpacked/diskpacked.go
tgulacsi: Done.

PR #1695, opened 2025-05-18
handle undefined image

quinn:
Some of the images I've loaded in have this attribute missing, which causes the home page to break. It could be an issue with webp images, i haven't investigated much. I'm loading them using `(*Client).UploadFile`. Jpegs seem to work fine.

### DIFF
--- a/server/perkeepd/ui/blob_item_image_content.js
+++ b/server/perkeepd/ui/blob_item_image_content.js
@@ -125,7 +125,7 @@ cam.BlobItemImageContent.getHandler = function(blobref, searchSession, href) {
 	var cci = cam.permanodeUtils.getSingleAttr(m.permanode, 'camliContentImage');
 	if (cci) {
 		var ccim = searchSession.getResolvedMeta(cci);
-		if (ccim) {
+		if (ccim && ccim.image) {
 			return new cam.BlobItemImageContent.Handler(ccim, href, searchSession.getTitle(blobref));
 		}
 	}


quinn:
I'd also be happy to try to fix the underlying issue with webp if someone can point me in the right direction, assuming thats the actual issue and i'm not doing something wrong with the uploader

quinn:
added `golang.org/x/image/webp` package

quinn:
Updated to add support for displaying svg images

pkg/server/image.go
```diff
@@ -338,61 +338,90 @@ var singleResize singleflight.Group
...
-	etag := blob.RefFromString(key).String()[5:]
-	inm := req.Header.Get("If-None-Match")
-	if inm != "" {
-		if strings.Trim(inm, `"`) == etag {
-			thumbCacheHeader304.Add(1)
-			rw.WriteHeader(http.StatusNotModified)
+	var imageData []byte
+	var etag string
+	if db.File.MIMEType == "image/svg+xml" {
+		etag = file.String()[5:]
```
bradfitz: btw I fixed this old (invalid) 5 assumption in a recent commit

pkg/server/image.go
```diff
@@ -426,10 +455,3 @@ func (ih *ImageHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, fil
 		}
 	}
 }
-
-func imageContentTypeOfFormat(format string) string {
-	if format == "jpeg" {
-		return "image/jpeg"
-	}
-	return "image/png"
```
bradfitz, line 434: I also deleted this recently :)

internal/images/images.go
```diff
@@ -386,12 +389,96 @@ func decodeHEIFConfig(rat io.ReaderAt) (image.Config, error) {
...
+}
+
 // DecodeConfig returns the image Config similarly to
 // the standard library's image.DecodeConfig with the
 // addition that it also checks for an EXIF orientation,
 // and sets the Width and Height as they would visibly
 // be after correcting for that orientation.
-func DecodeConfig(r io.Reader) (Config, error) {
+func DecodeConfig(mimetype string, r io.Reader) (Config, error) {
+	if mimetype == "image/svg+xml" {
```
bradfitz: do we really need this signature change? we're already reading the header & sniffing below. we could just call `isSVG` ourselves in this func if image.DecodeConfig fails. Then read a few KB, check isSVG, and then call dimsFromSVG with the stitched together `io.MultiReader(&buf, r)` like below?

bradfitz (commented):
Nice, thanks!

You want me to take this over, or do you want to make these tweaks?

internal/images/images.go
quinn: makes sense! It's a bit awkward having it passed in here, but wasn't sure the best way to handle it. I'll give that a try

quinn: Alright, I refactored this to not need the arg. 

Do you think i should load the whole `magic` package here, or make `isSVG` exported and use it directly?

pkg/server/image.go
quinn: can you point me to the commit? couldn't find anything relevant in recent commits on master

quinn: do you mean a potential panic for len < 5

bradfitz: Oh, I guess I haven't merged it yet: https://github.com/perkeep/perkeep/pull/1769

bradfitz: the len(5) was for len("sha1-") but nowadays SHA-1 isn't even in the default and the it should be sha224- instead.

So I changed it to strings.Cut on "-".

internal/images/images.go
bradfitz: Either way. Or maybe we should mimic the Go std style with the image package and its registration mechanism and have our own image type registration system with a richer API than https://pkg.go.dev/image#RegisterFormat and its magic format.

Then we could make our own svg package with all the code and have it register itself with the images (and magic?) package.

The advantage of that is that you could then use build tags to make certain image formats optional if they get too bloaty dep-wise. (e.g. HEIC support bringing in a wasm interpreter, or SVG bringing in XML decoding)

But we can do that later if/when binary size becomes a concern.

I'd just do whatever's easiest for now.

pkg/server/image.go
quinn: alright, made this change

internal/images/images.go
quinn: OK, what I have here is just importing the `magic` package, and it seems to be working. If that works for you, this should be ok to merge. 

I'd be down to do another PR after this one to refactor the magic package to use a registration system. 

in general i think extending image preview support would be good, i recently tried to get Go to display a preview image for every file i had in some backups: https://github.com/quinn/file-browser/blob/main/internal/server/preview.go

internal/images/images.go
```diff
@@ -386,6 +390,78 @@ func decodeHEIFConfig(rat io.ReaderAt) (image.Config, error) {
...
 
+// parseLeadingNumber strips any unit suffix and converts the leading
+// number to float64.  Returns ok=false if it can’t.
+func parseLeadingNumber(s string) (v float64, ok bool) {
+	s = strings.TrimSpace(s)
+	i := 0
+
+	for _, r := range s {
+		if unicode.IsDigit(r) || r == '+' || r == '-' || r == '.' {
+			i++
```
bradfitz, line 406: Note that "range s" is ranging over UTF-8 codepoints in a string, but then you're slicing by matched bytes on line 409, not by runes)

So this whole func fails for non-ASCII digits that encode to more than one UTF-8 byte: https://go.dev/play/p/UmbLf83PK98

I think you actually meant to avoid `unicode.IsDigit` here and check for `'0' <= r && r <= '9'` instead, or do an earlier check for `if r >= utf8.RuneSelf { break }`

internal/images/images.go
```diff
@@ -386,6 +390,78 @@ func decodeHEIFConfig(rat io.ReaderAt) (image.Config, error) {
...
+			case "height":
+				heightAttr = a.Value
+			case "viewBox":
+				viewBoxAttr = a.Value
+			}
+		}
+
+		// Try width & height first
+		if w, ok = parseLeadingNumber(widthAttr); ok {
+			if h, ok = parseLeadingNumber(heightAttr); ok {
```
bradfitz: I'd prefer `:=` instead of `=` for these two, to keep the scope the variables shorter and easier to reason about (so reader doesn't have to consider whether `w' matter before/after this chunk of code.

pkg/server/image.go
```diff
@@ -338,61 +338,90 @@ var singleResize singleflight.Group
...
+	var etag string
+	if db.File.MIMEType == "image/svg+xml" {
+		_, etag, _ = strings.Cut(file.String(), "-")
+		fr, err := schema.NewFileReader(ctx, ih.Fetcher, file)
+		if err != nil {
+			http.Error(rw, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		defer fr.Close()
+		imageData, err = io.ReadAll(fr)
```
bradfitz: This is fine for now but maybe not ideal, reading it all into memory just to serve copy it back out to the network.

Maybe leave a TODO comment to use https://pkg.go.dev/net/http#ServeContent on the schema.NewFileReader result directly, avoiding the ReadAll?

pkg/server/image.go
```diff
@@ -338,61 +338,90 @@ var singleResize singleflight.Group
...
 			return
 		}
 	} else {
-		if !disableThumbCache && req.Header.Get("If-Modified-Since") != "" {
-			thumbCacheHeader304.Add(1)
-			rw.WriteHeader(http.StatusNotModified)
+		if !httputil.IsGet(req) {
+			http.Error(rw, "Invalid method", http.StatusBadRequest)
+			return
+		}
```
bradfitz: shouldn't this be up earlier, before the mime type check? doesn't seem svg-vs-not specific?

pkg/server/image.go
bradfitz: Actually if you did that now (if it's not too onerous), you could return early and outdent the rest of the code of the non-svg codepath.

internal/images/images.go
```diff
@@ -386,6 +390,78 @@ func decodeHEIFConfig(rat io.ReaderAt) (image.Config, error) {
...
+	}
+	if i == 0 {
+		return 0, false
+	}
+	n, err := strconv.ParseFloat(s[:i], 64)
+	return n, err == nil
+}
+
+// dimsFromSVG reads tokens until it sees the first <svg>
+// start-element, then extracts width/height/viewBox.
```
bradfitz: the "viewBox" thing wasn't obvious to me (as somebody who doesn't know SVG) ... looked like a documentation mistake because the signature on the next line had width & height but no viewbox mentioned.

Maybe just say:

// dimsFromSVG reads XML tokens from SVG file r until it can compute
// the dimensions of the image (in "abstract user units").

bradfitz:
Looks pretty good to me. Last round of minor comments! Thanks for your patience.

pkg/server/image.go
quinn: I refactored this to use `ServeContent`, I'm not 100% sure about it, all the other feedback should be done. Let me know what you think! If it looks good to merge, I'll squash everything down again

internal/images/images.go
quinn: this one doesn't say "outdated", but I went with the `RuneSelf` check earlier in the loop

internal/images/images.go
```diff
@@ -397,22 +398,27 @@ func parseLeadingNumber(s string) (v float64, ok bool) {
...
 	n, err := strconv.ParseFloat(s[:i], 64)
 	return n, err == nil
 }
 
-// dimsFromSVG reads tokens until it sees the first <svg>
-// start-element, then extracts width/height/viewBox.
-func dimsFromSVG(r io.Reader) (w, h float64, err error) {
+// dimsFromSVG reads XML tokens from SVG file r until it can compute
+// the dimensions of the image (in "abstract user units").
+func dimsFromSVG(r io.Reader) (float64, float64, error) {
```
bradfitz: I would still name these results for documentation purposes.

It's okay to have the `w, ok :=` and `h, ok :=` variables below be different variables (with the same name in a shadowed scope)

bradfitz: I pushed this edit, and the squash.

pkg/server/image.go
```diff
@@ -337,19 +337,58 @@ func (ih *ImageHandler) scaleImage(ctx context.Context, fileRef blob.Ref) (*form
...
 	if !httputil.IsGet(req) {
 		http.Error(rw, "Invalid method", http.StatusBadRequest)
 		return
 	}
+
+	ctx := req.Context()
+	res, err := ih.Search.Describe(ctx, &search.DescribeRequest{
+		BlobRef: file,
+		Depth:   1,
+	})
```
bradfitz, line 373: one downside of this change is that even for If-None-Match requests where we previously replied "304 Not Modified!" right away without doing any I/O, now we're doing a heavy Describe request just to do nothing with the result.

pkg/server/image.go
```diff
@@ -337,19 +337,58 @@ func (ih *ImageHandler) scaleImage(ctx context.Context, fileRef blob.Ref) (*form
...
+		defer fr.Close()
+
+		h := rw.Header()
+		if !disableThumbCache {
+			h.Set("Expires", time.Now().UTC().Add(oneYear).Format(http.TimeFormat))
+			h.Set("Etag", strconv.Quote(etag))
+		}
+		h.Set("Content-Type", db.File.MIMEType)
+
+		if req.Method == "GET" {
```
bradfitz: I don't think you need (or want) this. For a HEAD request, you still want to set the Content-Length, etc. ServeContent will do that for you.

bradfitz:
I'll be back in an ~hour and can make those tweaks if you don't beat me to it.

pkg/server/image.go
bradfitz: Fixed.

pkg/server/image.go
bradfitz: Fixed.

quinn:
@bradfitz thanks for fixing/merging!

Issue #1697, opened 2025-05-19
Update to Go version 1.24.x


radkat:
Latest stable version is 1.24.3

PR #1698, opened 2025-05-19
all: Update Go version

radkat:
As a first pass, I only fixed the tests that started failing as a result of version update. 

Addresses the core of #1697


bradfitz (approved):

bradfitz:
Looks like CI is still using Go 1.21. Should bump that too.

internal/closure/jstest/jstest.go
```diff
@@ -127,6 +127,6 @@ func TestCwd(t *testing.T) {
 	c := exec.Command("mocha", "-r", bootstrap, "-r", depsFile, filepath.Join(path, "*test.js"))
 	b, err := c.CombinedOutput()
 	if err != nil {
-		t.Fatalf(string(b))
+		t.Fatalf("mocha failed: %v\n%s", err, string(b))
```
bradfitz, line 130: it doesn't matter in this case, but FYI with %s, both string and `[]byte` are handled the same way, so this `string(b)` is now redundant (and actually just allocates)

But we don't care about performance here.

bradfitz:
Let's squash those two commits together into one since they're related.

PR #1701, opened 2025-05-24
[draft] bluesky importer

### LINKED ISSUE #1694: Bluesky importer
"Can I get a copy of all of my data?" from https://bsky.social/about/blog/5-19-2023-user-faq -> https://docs.bsky.app/blog/repo-export

quinn:
fixes #1694


bradfitz:
This is exciting! I'm a Bluesky user myself and this has been on my list.

I'm happy to help work on this with you.

bradfitz:
> Bluesky has a very sophisticated API which has its own concept of content addressable storage, which needs to be backed by a KV storage, which I've implemented here as part of the importer, which is just an index of another CAS (perkeep). It works, but if there is a more clever way to have perkeep directly back their CAR, that could be cool.

I haven't worked on Bluesky stuff in a couple years (https://github.com/bluesky-social/indigo/commits?author=bradfitz was about as much as I ever did) so I need to remember how it all works. Plus it's probably changed a bit.

But ideally we'd map everything 1:1 as much as possible. Perkeep uses sha224 by default (for length), but we could add an opt-in sha256 blob ref type so the atproto values are stored directly under the same original hashes?

> do you have any feedback on the current implementation? If this generally looks ok, I think a logical next step could be to add tests, given the complexity

I haven't had a chance to look deeply yet.

pkg/blob/ref.go
```diff
@@ -212,6 +212,22 @@ func (r Ref) IsSupported() bool {
 	return ok
 }
 
+// FromBytes returns a ref object from a digest
+func FromBytes(digNameStr string, digest []byte) (Ref, error) {
```
quinn, line 216: @bradfitz I added this function, if there is a better (already existing) way to do this lmk

pkg/importer/bluesky/datastore.go
```diff
@@ -0,0 +1,151 @@
...
+	h, err := blob.NewHashOfType("sha256")
+	if err != nil {
+		return err
+	}
+
+	if _, err := h.Write(value); err != nil {
+		return err
+	}
+
+	if bytes.Compare(digest, h.Sum(nil)) != 0 {
```
quinn, line 108: This is probably not necessary since `Receive` will verify the digest

Issue #1705, opened 2025-09-27
Fix linter checks


radkat:
`golangci-lint` finds ~50 different issues in the existing codebase. 

Go over each category and fix if it makes sense.

PR #1706, opened 2025-09-27
internal/netutil: fix errcheck linter error

radkat:
Pay attention to the return error of `cmd.Wait()`

Part of working on #1705 

---

Was:

```
internal/netutil/ident.go:169:16: Error return value of `cmd.Wait` is not checked (errcheck)
internal/netutil/ident.go:175:24: Error return value of `cmd.Process.Kill` is not checked (errcheck)
internal/netutil/ident.go:206:16: Error return value of `cmd.Wait` is not checked (errcheck)
internal/netutil/ident.go:212:24: Error return value of `cmd.Process.Kill` is not checked (errcheck)
```

### DIFF
--- a/internal/netutil/ident.go
+++ b/internal/netutil/ident.go
@@ -41,7 +41,7 @@ var (
 )
 
 // ConnUserid returns the uid that owns the given localhost connection.
-// The returned error is ErrNotFound if the connection wasn't found.
+// If the connection is not found, it returns ErrNotFound.
 func ConnUserid(conn net.Conn) (uid int, err error) {
 	return AddrPairUserid(conn.LocalAddr(), conn.RemoteAddr())
 }
@@ -73,19 +73,18 @@ func HostPortToIP(hostport string, ctx *net.TCPAddr) (hostaddr *net.TCPAddr, err
 
 // AddrPairUserid returns the local userid who owns the TCP connection
 // given by the local and remote ip:port (lipport and ripport,
-// respectively).  Returns ErrNotFound for the error if the TCP connection
-// isn't found.
+// respectively). If the TCP connection is not found, it returns ErrNotFound.
 func AddrPairUserid(local, remote net.Addr) (uid int, err error) {
 	lAddr, lOk := local.(*net.TCPAddr)
 	rAddr, rOk := remote.(*net.TCPAddr)
 	if !(lOk && rOk) {
-		return -1, fmt.Errorf("netutil: Could not convert Addr to TCPAddr")
+		return 0, fmt.Errorf("netutil: Could not convert Addr to TCPAddr")
 	}
 
 	localv4 := (lAddr.IP.To4() != nil)
 	remotev4 := (rAddr.IP.To4() != nil)
 	if localv4 != remotev4 {
-		return -1, fmt.Errorf("netutil: address pairs of different families; localv4=%v, remotev4=%v",
+		return 0, fmt.Errorf("netutil: address pairs of different families; localv4=%v, remotev4=%v",
 			localv4, remotev4)
 	}
 
@@ -101,7 +100,7 @@ func AddrPairUserid(local, remote net.Addr) (uid int, err error) {
 		}
 		f, err := os.Open(file)
 		if err != nil {
-			return -1, fmt.Errorf("Error opening %s: %v", file, err)
+			return 0, fmt.Errorf("Error opening %s: %v", file, err)
 		}
 		defer f.Close()
 		return uidFromProcReader(lAddr.IP, lAddr.Port, rAddr.IP, rAddr.Port, f)
@@ -140,19 +139,19 @@ func uidFromUsernameFn(username string) (uid int, err error) {
 		return uid, nil
 	}
 	u, err := user.Lookup(username)
-	if err == nil {
-		uid, err := strconv.Atoi(u.Uid)
-		return uid, err
+	if err != nil {
+		return 0, err
 	}
-	return 0, err
+	return strconv.Atoi(u.Uid)
 }
 
 func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err error) {
 	seek := fmt.Sprintf("%s:%d->%s:%d", maybeBrackets(lip), lport, maybeBrackets(rip), rport)
 	seekb := []byte(seek)
-	if _, err = exec.LookPath("lsof"); err != nil {
-		return
+	if _, err := exec.LookPath("lsof"); err != nil {
+		return 0, err
 	}
+
 	cmd := exec.Command("lsof",
 		"-b", // avoid system calls that could block
 		"-w", // and don't warn about cases where -b fails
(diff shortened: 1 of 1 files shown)


internal/netutil/ident.go
```diff
@@ -198,20 +200,34 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
+	}
+
 	err = cmd.Start()
 	if err != nil {
-		return -1, err
+		return nil, err
 	}
-	defer cmd.Process.Kill()
 
-	return uidFromSockstatReader(lip, lport, rip, rport, stdout)
```
bradfitz, line 214: Start
uid, err := uidFromSockstatReader
Wait + err check
return uid, err

internal/netutil/ident.go
```diff
@@ -153,7 +157,11 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
 	if _, err = exec.LookPath("lsof"); err != nil {
 		return
 	}
-	cmd := exec.Command("lsof",
+
+	ctx, cancel := context.WithTimeout(context.Background(), cmdTimeout)
+	defer cancel()
```
bradfitz: remove

internal/netutil/ident.go
```diff
@@ -140,11 +140,10 @@ func uidFromUsernameFn(username string) (uid int, err error) {
...
 	}
 	u, err := user.Lookup(username)
-	if err == nil {
-		uid, err := strconv.Atoi(u.Uid)
-		return uid, err
+	if err != nil {
+		return 0, err
 	}
-	return 0, err
+	return strconv.Atoi(u.Uid)
```
radkat, line 145: I flipped this around just because otherwise it was unusual to read.

internal/netutil/ident.go
radkat: Done.

internal/netutil/ident.go
radkat: Done.

internal/netutil/ident.go
```diff
@@ -41,7 +41,7 @@ var (
 )
 
 // ConnUserid returns the uid that owns the given localhost connection.
-// The returned error is ErrNotFound if the connection wasn't found.
+// If the connection is not found, returns -1 and ErrNotFound.
```
bradfitz: This invites the question: for other types of errors, might it not return -1? In that case, the caller can't depend on this being safe anyway:

     uid, _ := netutil.ConnUserid(c)

In which case there's no point in the -1 guardrail for those ignoring errors anyway.

So we should either go all the way to Go-idiomatic (zero-value, non-nil-error) or if we really want the -1 guardrail, go all the way and say that on any error, it's -1. But that feels so weird to promise.

I think we should just abandon all negative one and be idiomatic and return zero values on errors.

internal/netutil/ident.go
```diff
@@ -162,17 +163,17 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
 	if err != nil {
 		return
 	}
-	defer cmd.Wait()
-	defer stdout.Close()
-	err = cmd.Start()
-	if err != nil {
+	if err = cmd.Start(); err != nil {
 		return
 	}
```
bradfitz: this func is too long (more than 2-3 lines) for naked returns. Use `err := cmd.Start()` here and `return err` explicitly so caller doesn't have to reason about the scope

Likewise below.

internal/netutil/ident.go
```diff
@@ -190,28 +191,44 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
-	defer cmd.Wait()
-	defer stdout.Close()
-	err = cmd.Start()
-	if err != nil {
-		return -1, err
+	if err = cmd.Start(); err != nil {
+		return
+	}
+
+	res, resErr := uidFromSockstatReader(lip, lport, rip, rport, stdout)
```
bradfitz: why "resErr"? just err seems fine.

internal/netutil/ident.go
radkat: Done.

internal/netutil/ident.go
radkat: Done.

internal/netutil/ident.go
radkat: Because I return it later, after `err = cmd.Wait()` where it gets reassigned.

internal/netutil/ident.go
```diff
@@ -41,8 +41,8 @@ var (
 )
 
 // ConnUserid returns the uid that owns the given localhost connection.
-// The returned error is ErrNotFound if the connection wasn't found.
-func ConnUserid(conn net.Conn) (uid int, err error) {
+// If the connection is not found, returns ErrNotFound.
```
bradfitz: , it returns

internal/netutil/ident.go
```diff
@@ -73,19 +73,18 @@ func HostPortToIP(hostport string, ctx *net.TCPAddr) (hostaddr *net.TCPAddr, err
 
 // AddrPairUserid returns the local userid who owns the TCP connection
 // given by the local and remote ip:port (lipport and ripport,
-// respectively).  Returns ErrNotFound for the error if the TCP connection
-// isn't found.
-func AddrPairUserid(local, remote net.Addr) (uid int, err error) {
+// respectively). If the TCP connection is not found, returns ErrNotFound.
```
bradfitz: , it returns

(complete sentences in Go docs, per https://go.dev/wiki/CodeReviewComments#comment-sentences)

internal/netutil/ident.go
```diff
@@ -190,35 +189,41 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
 			continue
 		}
-		username := string(f[2])
-		return uidFromUsername(username)
+		res, resErr = uidFromUsername(string(f[2]))
+		break
 	}
-	return -1, ErrNotFound
 
+	if err = cmd.Wait(); err != nil {
```
bradfitz: :=

internal/netutil/ident.go
```diff
@@ -162,25 +161,25 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
 		//"-a",  // AND the following together:
 		// "-u", strconv.Itoa(uid)  // just this uid
 		"-itcp") // we only care about TCP connections
+
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		return
+		return 0, err
 	}
-	defer cmd.Wait()
```
bradfitz, line 169: I'd do something like:

```
defer func() {
   if werr := cmd.Wait(); werr != nil && err == nil {
        err = werr
   }
}()
```

... otherwise cmd.Wait in your proposed changed takes priority over any error from `uidFromUsername` which is more interesting.

internal/netutil/ident.go
```diff
@@ -190,35 +189,41 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
 	if err != nil {
-		return -1, err
+		return 0, err
 	}
-	defer cmd.Wait()
-	defer stdout.Close()
-	err = cmd.Start()
-	if err != nil {
-		return -1, err
+	if err = cmd.Start(); err != nil {
```
bradfitz: :=

internal/netutil/ident.go
```diff
@@ -162,25 +161,31 @@ func uidFromLsof(lip net.IP, lport int, rip net.IP, rport int) (uid int, err err
...
+		}
+	}()
+
+	var res int
+	var resErr error
 	br := bufio.NewReader(stdout)
 	for {
 		line, err := br.ReadSlice('\n')
 		if err == io.EOF {
 			break
```
bradfitz: on EOF now you're returning (zero, zero)

bradfitz (approved):

bradfitz:
Tests passed. Merge away.

PR #1707, opened 2025-10-04
internal/pinentry: fix linter errors

radkat:
Updates #1705

### DIFF
--- a/internal/pinentry/pinentry.go
+++ b/internal/pinentry/pinentry.go
@@ -75,8 +75,14 @@ func (r *Request) GetPIN() (pin string, outerr error) {
 	stdin, _ := cmd.StdinPipe()
 	stdout, _ := cmd.StdoutPipe()
 	check(cmd.Start())
-	defer cmd.Wait()
+
+	defer func() {
+		if werr := cmd.Wait(); werr != nil && outerr == nil {
+			outerr = werr
+		}
+	}()
 	defer stdin.Close()
+
 	br := bufio.NewReader(stdout)
 	lineb, _, err := br.ReadLine()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/pinentry/pinentry.go
```diff
@@ -67,8 +67,11 @@ func (r *Request) GetPIN() (pin string, outerr error) {
 	stdin, _ := cmd.StdinPipe()
 	stdout, _ := cmd.StdoutPipe()
 	check(cmd.Start())
-	defer cmd.Wait()
-	defer stdin.Close()
```
bradfitz, line 79: why'd you remove the stdin.Close? Does it exit at all if you don't close this?

radkat: Doc say "The pipe will be closed automatically after [Cmd.Wait](https://pkg.go.dev/os/exec#Cmd.Wait) sees the command exit. A caller need only call Close to force the pipe to close sooner. For example, if the command being run will not exit until standard input is closed, the caller must close the pipe." -- is this the case here?

bradfitz: yes, that's the case

internal/pinentry/pinentry.go
```diff
@@ -60,23 +59,26 @@ func check(err error) {
...
+	defer func() {
+		if werr := cmd.Wait(); werr != nil && outerr == nil {
+			outerr = werr
+		}
+	}()
+	defer func() {
+		if cerr := stdin.Close(); cerr != nil && outerr == nil {
+			outerr = cerr
+		}
+	}()
```
bradfitz: revert this to just be a `defer stdin.Close()` .... it can't return an error really

PR #1714, opened 2025-10-04
pkg/cmdmain: fix linter errors

radkat:
Updates #1705

### DIFF
--- a/pkg/cmdmain/cmdmain.go
+++ b/pkg/cmdmain/cmdmain.go
@@ -392,10 +392,7 @@ var sysExec func(argv0 string, argv []string, envv []string) (err error)
 // for it to finish.
 func runExec(bin string, args []string, e *env) error {
 	if sysExec != nil {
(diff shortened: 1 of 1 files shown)


pkg/cmdmain/cmdmain.go
```diff
@@ -392,7 +392,10 @@ var sysExec func(argv0 string, argv []string, envv []string) (err error)
 // for it to finish.
 func runExec(bin string, args []string, e *env) error {
 	if sysExec != nil {
-		sysExec(bin, append([]string{filepath.Base(bin)}, args...), e.flat())
+		err := sysExec(bin, append([]string{filepath.Base(bin)}, args...), e.flat())
+		if err != nil {
+			return err
+		}
```
bradfitz: syscall.Exec can never return a nil error, so you can just "return sysExec(..." here

bradfitz: (as upon a successful exec, exec never returns at all, replacing the process)

bradfitz (approved):

PR #1715, opened 2025-10-04
internal/osutil: fix linter errors

radkat:
Updates #1705

### DIFF
--- a/internal/osutil/mem_unix.go
+++ b/internal/osutil/mem_unix.go
@@ -29,7 +29,7 @@ func init() {
 
 func memUnix() int64 {
 	var ru syscall.Rusage
-	syscall.Getrusage(0, &ru)
+	syscall.Getrusage(0, &ru) // nolint:errcheck
 	if runtime.GOOS == "linux" {
 		// in KB
 		return int64(ru.Maxrss) << 10
--- a/internal/osutil/paths.go
+++ b/internal/osutil/paths.go
@@ -102,7 +102,10 @@ func CamliVarDir() (string, error) {
 	newName := camliVarDirOf("perkeep")
 
 	if fi, err := os.Lstat(oldName); err == nil && fi.IsDir() && oldName != newName {
-		n := numRegularFilesUnder(oldName)
+		n, err := numRegularFilesUnder(oldName)
+		if err != nil {
+			return "", fmt.Errorf("error enumerating old directory %s: %w", oldName, err)
+		}
 		if n == 0 {
 			log.Printf("removing old, empty var directory %s", oldName)
 			os.RemoveAll(oldName)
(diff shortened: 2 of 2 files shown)


internal/osutil/paths.go
```diff
@@ -114,6 +114,7 @@ func CamliVarDir() (string, error) {
 }
 
 func numRegularFilesUnder(dir string) (n int) {
+	// nolint:errcheck
 	filepath.Walk(dir, func(path string, fi os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
bradfitz: Yeah, considering the caller uses this to os.RemoveAll a directory, changing its signature to return an error probably makes sense, and then the caller can check it.

internal/osutil/mem_unix.go
```diff
@@ -29,7 +29,7 @@ func init() {
 
 func memUnix() int64 {
 	var ru syscall.Rusage
-	syscall.Getrusage(0, &ru)
+	syscall.Getrusage(0, &ru) // nolint:errcheck
```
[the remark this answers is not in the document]
bradfitz: meh. doesn't matter. You can ignore Copilot here.

internal/osutil/paths.go
```diff
@@ -113,8 +116,8 @@ func CamliVarDir() (string, error) {
 	return newName, nil
 }
 
-func numRegularFilesUnder(dir string) (n int) {
-	filepath.Walk(dir, func(path string, fi os.FileInfo, err error) error {
+func numRegularFilesUnder(dir string) (n int, err error) {
+	err = filepath.Walk(dir, func(path string, fi os.FileInfo, err error) error {
```
bradfitz, line 120: I'd do "return filepath.Walk" here

internal/osutil/paths.go
```diff
@@ -113,8 +116,8 @@ func CamliVarDir() (string, error) {
...
 		if fi != nil && fi.Mode().IsRegular() {
```
bradfitz, line 124: and add:

    if err != nil {
         return err
    }

... inside this func.

internal/osutil/paths.go
```diff
@@ -171,7 +174,10 @@ func perkeepConfigDir() (string, error) {
 	oldName := configDirNamed("camlistore")
 	newName := configDirNamed("perkeep")
 	if fi, err := os.Lstat(oldName); err == nil && fi.IsDir() && oldName != newName {
-		n := numRegularFilesUnder(oldName)
+		n, err := numRegularFilesUnder(oldName)
+		if err != nil {
+			return "", err
```
bradfitz: return "", fmt.Errorf("error enumerating old directory %s: %w", oldName, err)

internal/osutil/paths.go
radkat: The signature requires two things to be returned so I can do that.

internal/osutil/paths.go
```diff
@@ -113,14 +116,17 @@ func CamliVarDir() (string, error) {
...
+			return err
+		}
 		if fi != nil && fi.Mode().IsRegular() {
 			n++
 		}
 		return nil
 	})
-	return
+	return n, err
 }
```
radkat, line 130: Per our chat, reverting back to a variable so that it's deterministic behavior.

Issue #1717, opened 2025-10-07
Remove scary red warning


radkat:
Right now when accessing a page via ts..net, Perkeep adds a red text saying "This page is not served securely (no https). Proceed at your own risk." -- remove it for this case.

PR #1718, opened 2025-10-08
all: remove red banner when Tailscale

radkat:
#1717


pkg/importer/importer.go
```diff
@@ -199,6 +199,7 @@ func init() {
 // HostConfig holds the parameters to set up a Host.
 type HostConfig struct {
 	BaseURL      string
+	RequireHTTPS bool
```
bradfitz: let's instead thread this all the way through the http.Request.Context like

```patch
diff --git a/pkg/webserver/webserver.go b/pkg/webserver/webserver.go
index 3fc11fe02..7e2eccb36 100644
--- a/pkg/webserver/webserver.go
+++ b/pkg/webserver/webserver.go
@@ -178,6 +178,12 @@ func (tw *trackResponseWriter) Write(p []byte) (int, error) {
        return tw.ResponseWriter.Write(p)
 }
 
+func (s *Server) isTailscale() bool {
+       preColon, _, _ := strings.Cut(addr, ":")
+       isTailscale := preColon == "tailscale"
+
+}
+
 // Listen starts listening on the given host:port addr.
 //
 // If the "host" part is "tailscale", it goes into Tailscale tsnet mode, and the
@@ -192,11 +198,8 @@ func (s *Server) Listen(addr string) error {
                return fmt.Errorf("<host>:<port> needs to be provided to start listening")
        }
 
-       preColon, _, _ := strings.Cut(addr, ":")
-       isTailscale := preColon == "tailscale"
-
        var err error
-       if isTailscale {
+       if s.isTailscale() {
                s.listener, err = s.listenTailscale(addr, s.enableTLS)
        } else {
                s.listener, err = listen.Listen(addr)
@@ -339,6 +342,13 @@ func (s *Server) Serve() {
 
        srv := &http.Server{
                Handler: s,
+               BaseContext: func(net.Listener) context.Context {
+                       ctx := context.Background()
+                       if s.isTailscale() {
+                               ctx = context.WithValue(...)
+                       }
+                       return ctx
+               },
        }
        // TODO: allow configuring src.ErrorLog (and plumb through to
        // Google Cloud Logging when run on GCE, eventually)
```

pkg/importer/importer.go
```diff
@@ -846,8 +856,11 @@ func (im *importer) ShowClientAuthEditForm() bool {
 	return im.props.NeedsAPIKey
 }
 
-func (im *importer) InsecureForm() bool {
-	return !strings.HasPrefix(im.host.ImporterBaseURL(), "https://")
+func (im *importer) SecureForm() bool {
+	if im.host.requireHTTPS {
+		return strings.HasPrefix(im.host.ImporterBaseURL(), "https://")
+	}
+	return true
```
bradfitz: instead of a method on importer, pass down "IsSecureTransport bool" to the template data and then make a func like

```go
func isSecureTransport(r *isSecureTransport) bool {
```

where you check either `r.TLS != nil` or the context value from above

pkg/webserver/webserver.go
```diff
@@ -339,6 +343,14 @@ func (s *Server) Serve() {
 
 	srv := &http.Server{
 		Handler: s,
+		BaseContext: func(l net.Listener) context.Context {
```
bradfitz: ln, never l

pkg/webserver/webserver.go
```diff
@@ -339,6 +343,14 @@ func (s *Server) Serve() {
...
+			ctx := context.Background()
+			if isTailscale(s.ListenURL()) {
+				// TODO(radkat): convert this to a custom context type (but where to put its definition?)
+				ctx = context.WithValue(ctx, "tailscale", true)
```
bradfitz: we should use https://pkg.go.dev/tailscale.com/util/ctxkey since we're importing the tailscale.com module anyway

pkg/importer/html.go
```diff
@@ -128,7 +129,7 @@ var tmpl = template.Must(template.New("root").Funcs(map[string]interface{}{
 </ul>
 
 {{if .Importer.ShowClientAuthEditForm}}
-	{{if .Importer.InsecureForm}}
+	{{if .SecureTransport or not .Importer.SecureForm}}
```
bradfitz: `{{if .ShowInsecureWarning}}` and then push policy up to Go, keeping the template dumber so policy isn't spread out over two places

pkg/webserver/webserver.go
radkat: I had to bump the Tailscale library version to be able to use that (and that led to a chain reaction).

PR #1725, opened 2025-10-21
pkg/importer: fix linter errors

radkat:
Updates #1705

### DIFF
--- a/pkg/importer/html.go
+++ b/pkg/importer/html.go
@@ -35,7 +35,7 @@ func (h *Host) execTemplate(w http.ResponseWriter, r *http.Request, data any) {
 		http.Error(w, fmt.Sprintf("Error executing template %q: %v", tmplName, err), 500)
 		return
 	}
-	w.Write(buf.Bytes())
+	w.Write(buf.Bytes()) // nolint:errcheck
 }
 
 type importersRootPage struct {
--- a/pkg/importer/importer.go
+++ b/pkg/importer/importer.go
@@ -237,7 +237,7 @@ func NewHost(hc HostConfig) (*Host, error) {
 		h.importers = append(h.importers, k)
 		clientId, clientSecret := hc.ClientId[k], hc.ClientSecret[k]
 		if clientSecret != "" && clientId == "" {
-			return nil, fmt.Errorf("Invalid static configuration for importer %q: clientSecret specified without clientId", k)
+			return nil, fmt.Errorf("invalid static configuration for importer %q: clientSecret specified without clientId", k)
 		}
 		props := impl.Properties()
 		imp := &importer{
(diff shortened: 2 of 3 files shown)


pkg/importer/noop.go
```diff
@@ -43,14 +43,14 @@ func (*todoImp) SummarizeAccount(acctNode *Object) string { return "" }
...
 		ctx.CallbackURL(),
 		"ctx.AccountURL()")
-	return nil
+	return err
 }
 
 func (*todoImp) ServeCallback(w http.ResponseWriter, r *http.Request, ctx *SetupContext) {
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-	fmt.Fprintf(w, "The callback page for the TODO importer.\n")
+	fmt.Fprintf(w, "The callback page for the TODO importer.\n") // nolint:errcheck
```
bradfitz, line 55: I don't want these annotations all over for things like read-only os.File.Close or ResponseWriter.Write calls.

pkg/importer/noop.go
```diff
@@ -43,14 +43,14 @@ func (*todoImp) SummarizeAccount(acctNode *Object) string { return "" }
 
 func (*todoImp) ServeSetup(w http.ResponseWriter, r *http.Request, ctx *SetupContext) error {
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-	fmt.Fprintf(w, "The Setup page for the TODO importer.\nnode = %v\ncallback = %s\naccount URL = %s\n",
+	_, err := fmt.Fprintf(w, "The Setup page for the TODO importer.\nnode = %v\ncallback = %s\naccount URL = %s\n",
 		ctx.AccountNode,
 		ctx.CallbackURL(),
 		"ctx.AccountURL()")
-	return nil
+	return err
```
bradfitz, line 50: revert; we don't care about ResponseWriter.Write calls

pkg/importer/importer.go
```diff
@@ -619,7 +619,7 @@ func (h *Host) serveImporterAcctCallback(w http.ResponseWriter, r *http.Request,
 		return
 	}
 	if !acctRef.Valid() {
-		httputil.ServeError(w, r, errors.New("No valid blobref returned from CallbackRequestAccount(r)"))
+		httputil.ServeError(w, r, errors.New("no valid blobref returned from CallbackRequestAccount(r)"))
```
bradfitz, line 622: these case ones are fine

pkg/importer/html.go
```diff
@@ -35,7 +35,7 @@ func (h *Host) execTemplate(w http.ResponseWriter, r *http.Request, data any) {
 		http.Error(w, fmt.Sprintf("Error executing template %q: %v", tmplName, err), 500)
 		return
 	}
-	w.Write(buf.Bytes())
+	w.Write(buf.Bytes()) // nolint:errcheck
```
bradfitz, line 38: bytes.Buffer writes can't fail

PR #1729, opened 2025-10-23
pkg/schema: fix linter errors

radkat:
Updates #1705

### DIFF
--- a/pkg/schema/blob.go
+++ b/pkg/schema/blob.go
@@ -316,7 +316,7 @@ type StaticSymlink struct {
 // non-empty. Otherwise it returns the contents of symlinkTargetBytes
 // concatenated as a string.
 func (sl StaticSymlink) SymlinkTargetString() string {
-	return sl.StaticFile.b.ss.SymlinkTargetString()
+	return sl.b.ss.SymlinkTargetString()
 }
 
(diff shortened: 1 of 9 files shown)


pkg/schema/dirreader.go
```diff
@@ -76,7 +76,11 @@ func (ss *superset) setFromBlobRef(ctx context.Context, fetcher blob.Fetcher, bl
 	if err != nil {
 		return fmt.Errorf("schema/dirreader: fetching schema blob %s: %v", blobRef, err)
 	}
-	defer rc.Close()
+	defer func() {
+		if cerr := rc.Close(); err == nil {
+			err = cerr
+		}
+	}()
```
bradfitz: this is overkill. Close on read-only FDs don't return errors. It only matters to check Close errors on things opened for write

pkg/schema/blob.go
```diff
@@ -432,7 +432,10 @@ func (bb *Builder) Blob() *Blob {
 		panic(err)
 	}
 	h := blob.NewHash()
-	io.WriteString(h, json)
+	_, err = io.WriteString(h, json)
+	if err != nil {
+		panic(err)
+	}
```
bradfitz: these writes can't fail so this is kinda just a distraction

PR #1732, opened 2025-10-29
pkg/importer: allow to skip paid API calls

### LINKED ISSUE #1731: Swarm API serves 402 for downloading images
Swarm importer is spitting out errors like

```
2025/10/24 03:11:29 Foursquare importer: error importing photos for checkin XXX: Get request on https://api.foursquare.com/v2/<reducted> failed with: 402 Payment Required
```

-- we need to back off -- or don't even try after having a first such error.

radkat:
Fixes #1731

### DIFF
--- a/pkg/importer/importer.go
+++ b/pkg/importer/importer.go
@@ -124,6 +124,13 @@ type Properties struct {
 	// For example, this is "foursquare" for "swarm", so "swarm" shows
 	// in the UI and URLs, but it's "foursquare" in permanodes.
 	PermanodeImporterType string
+
+	// HasSomePaidAPI specifies whether this service has part of
+	// its API requiring payment or quota for optional bits of data
+	// (such as Swarm venue photos).
+	// Whether or not to use such paid API(s) is determined for each
+	// run of each importer's account by the user.
+	HasSomePaidAPI bool
 }
 
 // LongPoller is optionally implemented by importers which can long
@@ -340,6 +347,11 @@ type RunContext struct {
 
 	mu           sync.Mutex // guards following
 	lastProgress *ProgressMessage
+}
+
+// UsePaidAPI reports whether the run is going to use the paid API(s) when retrieving data.
+func (rc *RunContext) UsePaidAPI() bool {
+	return rc.ia.UsePaidAPI()
 }
 
 // Context returns the run's context. It is always non-nil.
@@ -1132,6 +1144,7 @@ type importerAcct struct {
 	lastRunErr   error
 	lastRunStart time.Time
 	lastRunDone  time.Time
+	usePaidAPI   bool // whether to use the paid API(s) when retrieving data
 }
 
 func (ia *importerAcct) String() string {
@@ -1188,6 +1201,11 @@ func (ia *importerAcct) RefreshInterval() time.Duration {
 	return d
 }
 
+// UsePaidAPI reports whether the importer is going to use the paid API(s) when retrieving data.
+func (ia *importerAcct) UsePaidAPI() bool {
+	return ia.usePaidAPI
+}
+
 func (ia *importerAcct) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if r.Method == "POST" {
 		ia.serveHTTPPost(w, r)
@@ -1199,6 +1217,11 @@ func (ia *importerAcct) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Acct:     ia,
 		AcctType: fmt.Sprintf("%T", ia.im.impl),
 	}
+
+	if ia.im.props.HasSomePaidAPI {
+		body.HasSomePaidAPI = true
+	}
+
 	if run := ia.current; run != nil {
 		body.Running = true
 		body.StartedAgo = time.Since(ia.lastRunStart)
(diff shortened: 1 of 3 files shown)


pkg/importer/importer.go
```diff
@@ -342,6 +342,10 @@ type RunContext struct {
 	lastProgress *ProgressMessage
 }
 
+func (rc *RunContext) UsePaidAPI() bool {
```
bradfitz, line 353: document this

pkg/importer/importer.go
```diff
@@ -1132,6 +1136,7 @@ type importerAcct struct {
 	lastRunErr   error
 	lastRunStart time.Time
 	lastRunDone  time.Time
+	usePaidAPI   bool // whether to use the paid API(s) when retrieving data
```
bradfitz, line 1147: whether to use API endpoints (if any) that may require paying/credits

pkg/importer/swarm/swarm.go
```diff
@@ -282,10 +289,21 @@ func (r *run) importCheckins() error {
...
 			}
 
-			err = r.importPhotos(placeNode, dup)
-			if err != nil {
-				r.errorf("Foursquare importer: error importing photos for checkin %s: %v", checkin.Id, err)
-				continue
+			if r.RunContext.UsePaidAPI() {
+				err = r.importPhotos(placeNode, dup)
+				if err != nil {
+					isPaidErr := strings.Contains(err.Error(), "402 Payment Required")
```
bradfitz: should only check the status code here, not the undefined english

bradfitz: See https://go.dev/blog/go1.13-errors for %w etc

pkg/importer/importer.go
radkat: Now there's a separate bool (on the higher level, of the importer) that specifies whether or not there are paid APIs. This boolean now controls whether or not to use them.

pkg/importer/importer.go
```diff
@@ -124,6 +124,12 @@ type Properties struct {
 	// For example, this is "foursquare" for "swarm", so "swarm" shows
 	// in the UI and URLs, but it's "foursquare" in permanodes.
 	PermanodeImporterType string
+
+	// SupportsPaidAPI specifies whether this service has paid API(s)
+	// that we might need to use to extract data.
+	// Whether or not to use such paid API(s) is determined for each
+	// run of each importer's account by the user.
+	SupportsPaidAPI bool
```
bradfitz: how about "HasPaidAPI" 

Or really, this is about APIs that are mixed paid and non-paid, right? If it were just all paid, you wouldn't add it at all as an importer if you hadn't paid to get an API key.

So is this "HasSomePaidAPI" ?

pkg/importer/importer.go
```diff
@@ -342,6 +348,12 @@ type RunContext struct {
 	lastProgress *ProgressMessage
 }
 
+// UsePaidAPI returns `true` if the run is going to use the paid API(s) when retrieving data.
```
bradfitz: The typical Go way to document boolean functions is "reports whether" instead of "returns true if"

bradfitz: and instead of "use the paid API(s)" let's say "use any optional paid APIs when"

pkg/importer/importer.go
```diff
@@ -1188,6 +1201,12 @@ func (ia *importerAcct) RefreshInterval() time.Duration {
 	return d
 }
 
+// UsePaidAPI returns `true` if the importer is going to use the paid API(s) when retrieving data.
```
bradfitz: reports whether

pkg/importer/importer.go
```diff
@@ -1188,6 +1201,12 @@ func (ia *importerAcct) RefreshInterval() time.Duration {
...
+// Returns `false` otherwise.
```
bradfitz: can remove this line. But in general Go uses complete sentences ("It returns false otherwise.")

pkg/importer/swarm/swarm.go
```diff
@@ -282,10 +293,20 @@ func (r *run) importCheckins() error {
...
-			err = r.importPhotos(placeNode, dup)
-			if err != nil {
-				r.errorf("Foursquare importer: error importing photos for checkin %s: %v", checkin.Id, err)
-				continue
+			if r.RunContext.UsePaidAPI() {
+				err = r.importPhotos(placeNode, dup)
+				if err != nil {
+					isPaidErr := errors.Is(err, ErrPaymentRequired)
+					if isPaidErr {
+						if !showedPaidErrOnce {
```
bradfitz: In the case where we did already warn and it's a paid error, shouldn't we stop importing photos on future runs? That is, line 296 should be:

if r.RunContext.UsePaidAPI() && !showedPaidErrOnce {

?

radkat: Yeah, that's a good point. Prevent unnecessary calls.

pkg/importer/importer.go
```diff
@@ -124,6 +124,12 @@ type Properties struct {
 	// For example, this is "foursquare" for "swarm", so "swarm" shows
 	// in the UI and URLs, but it's "foursquare" in permanodes.
 	PermanodeImporterType string
+
+	// HasSomePaidAPI specifies whether this service has paid API(s)
+	// that we might need to use to extract data.
```
bradfitz: has part of its API requiring payment or quota for optional bits of data (such as Swarm venue photos).

bradfitz (approved):

PR #1746, opened 2025-11-13
pk-web: unescape title after md2html conversion

### LINKED ISSUE #1745: Unescape page title after md->html conversion
Running locally website with https://github.com/perkeep/perkeep/pull/1744 I see that the title of the page is `<title>Perkeep 0.12 (&amp;ldquo;Toronto&amp;rdquo;) - Perkeep</title>`

We're using https://github.com/russross/blackfriday to convert markdown to HTML -- I have a small PR unescaping the title specifically.

radkat:
Fixes #1745

### DIFF
--- a/website/pk-web/pkweb.go
+++ b/website/pk-web/pkweb.go
@@ -153,8 +153,8 @@ type pageParams struct {
 
 // pageTmplData is the template data passed to page.html.
 type pageTmplData struct {
-	Title    string
-	Subtitle string
+	Title    template.HTML
+	Subtitle template.HTML
 	Content  template.HTML
 
 	// For the "go-import" meta header:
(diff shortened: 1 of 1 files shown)


website/pk-web/pkweb.go
```diff
@@ -411,7 +412,7 @@ func serveFile(w http.ResponseWriter, r *http.Request, absPath string) {
 
 	title := ""
 	if m := h1TitlePattern.FindSubmatch(data); len(m) > 1 {
-		title = string(m[1])
+		title = html.UnescapeString(string(m[1]))
```
bradfitz: I suspect the real problem is `pageTmplData.Title` and `.SubTitle`'s types. Note how `Content` is already `template.HTML` so it's not auto-escaped by the html/template package. We should probably do the same for the title fields.

bradfitz (approved):

PR #1749, opened 2025-11-14
all: run modernize tool to simplify code

lotusirous:
As title, this CL runs the modernize 

```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
```

### DIFF
--- a/app/publisher/main.go
+++ b/app/publisher/main.go
@@ -31,6 +31,7 @@ import (
 	"io"
 	"io/fs"
 	"log"
+	"maps"
 	"net/http"
 	"net/url"
 	"os"
@@ -389,7 +390,7 @@ func newPublishHandler(conf *config) *publishHandler {
 		if err := os.MkdirAll(thumbsCacheDir, 0700); err != nil {
 			logger.Fatalf("Could not create cache dir %s for %v publisher: %v", thumbsCacheDir, conf.RootName, err)
 		}
-		kv, err := sorted.NewKeyValue(map[string]interface{}{
+		kv, err := sorted.NewKeyValue(map[string]any{
 			"type": "kv",
 			"file": filepath.Join(thumbsCacheDir, conf.RootName+"-thumbnails.kv"),
 		})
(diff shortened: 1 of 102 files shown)


bradfitz:
Almost all the omitempty removals should change to `omitzero`, really:

```patch
diff --git a/pkg/types/camtypes/discovery.go b/pkg/types/camtypes/discovery.go
index 1de0e0710..e1f8272a0 100644
--- a/pkg/types/camtypes/discovery.go
+++ b/pkg/types/camtypes/discovery.go
@@ -40,7 +40,7 @@ type Discovery struct {
        // StorageGenerationError is the error that occurred on generating the storage, if any.
        StorageGenerationError string `json:"storageGenerationError,omitempty"`
        // StorageInitTime is the initialization time of the storage.
-       StorageInitTime types.Time3339 `json:"storageInitTime,omitempty"`
+       StorageInitTime types.Time3339 `json:"storageInitTime"`

        ThumbVersion string `json:"thumbVersion"` // Thumbnailing version.

@@ -68,7 +68,7 @@ type SignDiscovery struct {
        // PublicKey is the path to the public signing key.
        PublicKey string `json:"publicKey,omitempty"`
        // PublicKeyBlobRef is the blob.Ref for the public key.
-       PublicKeyBlobRef blob.Ref `json:"publicKeyBlobRef,omitempty"`
+       PublicKeyBlobRef blob.Ref `json:"publicKeyBlobRef"`
        // PublicKeyID is the ID of the public key.
        PublicKeyID string `json:"publicKeyId"`
        // SignHandler is the URL path prefix to the signing handler.
diff --git a/pkg/types/camtypes/search.go b/pkg/types/camtypes/search.go
index 0b3d888c8..a0c29d234 100644
--- a/pkg/types/camtypes/search.go
+++ b/pkg/types/camtypes/search.go
@@ -114,7 +114,7 @@ type FileInfo struct {
        // WholeRef is the digest of the entire file contents.
        // This will be zero for non-regular files, and may also be zero
        // for files above a certain size threshold.
-       WholeRef blob.Ref `json:"wholeRef,omitempty"`
+       WholeRef blob.Ref `json:"wholeRef"`
 }
```

Can you either omit those from the diff, or fix all the omitempty issues in this PR, or a manual PR first, followed by rebasing this automated PR after?

lotusirous:
I updated. Thanks

PR #1755, opened 2025-11-26
internal/httputil: don't warn about running as root if in a container

### LINKED ISSUE #1752: Container is configured to run as root, but logs complain that it's running as root
the https://github.com/perkeep/perkeep/blob/master/Dockerfile doesn't add any user, so it is expecting to run s root.

But the `IsLocalhost` check is not happy about that: https://github.com/perkeep/perkeep/blob/545a90ced6f8b18f029c585d4a70d210592f91c3/internal/httputil/auth.go#L60

the check should probably be updated to detect if it's running in a container  and allow root in that case

bradfitz:
Fixes #1752

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -19,6 +19,7 @@ package env // import "perkeep.org/pkg/env"
 
 import (
 	"os"
+	"runtime"
 	"strconv"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


pkg/env/env.go
```diff
@@ -49,6 +50,30 @@ func OnGCE() bool {
...
+	if _, err := os.Stat("/run/.containerenv"); err == nil {
+		// See https://github.com/cri-o/cri-o/issues/5461
+		return true
+	}
+	v, _ := strconv.ParseBool(os.Getenv("PK_IN_CONTAINER"))
+	return v
+}
+
 var (
 	gceOnce sync.Once
```
radkat, line 78: I guess we should also use sync.OnceValue here

pkg/env/env.go
```diff
@@ -49,6 +50,30 @@ func OnGCE() bool {
...
 }
 
+// InContainer reports whether it seems like we're running in a container.
+func InContainer() bool {
+	return inContainerLazy()
+}
+
+var inContainerLazy = sync.OnceValue(inContainer)
+
+func inContainer() bool {
```
ptxmac, line 60: add `os.Getenv("KUBERNETES_SERVICE_HOST")` for detecting k8s

PR #1757, opened 2025-11-26
.github/workflows: add "dev" tag for any commit to master

radkat:


.github/workflows/docker.yml
```diff
@@ -52,6 +52,8 @@ jobs:
           tags: |
             # exact git tag (e.g., v1.2.3)
             type=ref,event=tag
+            # latest tag for master branch
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' && github.event_name == 'push' }}
```
bradfitz: let's not use "latest" as that normally means the latest released version. how about HEAD all caps? (if we're going to move from "master" to "main" one day)

bradfitz: actually apparently there's two entirely different schools of thought on whether "latest" means HEAD or "latest stable tagged release". cool 🙄

bradfitz: let's go with `value=dev` then and reserve `latest` for stable releases later.

PR #1778, opened 2025-12-30
pkg/importer/raindrop: fix panic in tests in CI

bradfitz:
We can't just use "client.New" in a package-level init: that uses the
user's client config files, but the server may not have a client
configured for the user it's running as.

Instead, we need to upload blobs to the importer's Host.

### DIFF
--- a/dev/devcam/test.go
+++ b/dev/devcam/test.go
@@ -148,6 +148,7 @@ func (c *testCmd) runTests(args []string) error {
 	}
 	env := c.env()
 	env.Set("SKIP_DEP_TESTS", "1")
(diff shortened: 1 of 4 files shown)


bradfitz:
cc @quinn

dev/devcam/test.go
```diff
@@ -148,6 +148,7 @@ func (c *testCmd) runTests(args []string) error {
 	}
 	env := c.env()
 	env.Set("SKIP_DEP_TESTS", "1")
+	env.Set("CAMLI_DISABLE_CLIENT_CONFIG_FILE", "1")
```
bradfitz, line 151: This makes the CI failure reproducible locally on a machine that has client config files available.

PR #1779, opened 2025-12-30
[wip] fix raindrop importer

quinn:
I've refactored this to not need the client at all. I don't actually need to store the json response, i'm just a pack rat 🐁 

when i tried this at first, the client was nil. I tried moving `targetStorage` to where i thought `target` would be available, then running the importer gave me this:

```
cannot discover if server has legacy sha1: client not using HTTP
```

There's probably something fundamental I don't get about how this is supposed to work, and it probably makes sense to not do client operations from within the server code anyway. I could refactor to save the json using `WriteFileFromReader` as well, but felt weird at the time (iirc) given it is not a "file" but just some data i have that would be nice to store.

### DIFF
--- a/pkg/importer/importer.go
+++ b/pkg/importer/importer.go
@@ -534,10 +534,10 @@ func (h *Host) InitHandler(hl blobserver.FindHandlerByTyper) error {
 		var err error
 		h.targetClient, err = client.New(client.OptionUseStorageClient(h.targetStorage))
 		if err != nil {
-			log.Printf("Raindrop importer: error creating client: %v", err)
+			log.Printf("importer: error creating client: %v", err)
 		}
 	} else {
-		log.Printf("Raindrop importer: Host does not have client")
+		log.Printf("importer: Host does not have client")
 	}
 
 	_, handler, _ = hl.FindHandlerByType("jsonsign")


pkg/importer/importer.go
```diff
@@ -533,6 +529,17 @@ func (h *Host) InitHandler(hl blobserver.FindHandlerByTyper) error {
...
+	h.targetStorage, _ = h.target.(blobserver.Storage)
+	if h.targetStorage != nil {
+		var err error
+		h.targetClient, err = client.New(client.OptionUseStorageClient(h.targetStorage))
+		if err != nil {
+			log.Printf("Raindrop importer: error creating client: %v", err)
+		}
+	} else {
+		log.Printf("Raindrop importer: Host does not have client")
+	}
```
quinn, line 541: @bradfitz looking for feedback on this, i thought this would make the client available, but it seems it is not configured properly / is completely the wrong object, or something. Also, since I refactored to not need the client anymore, I can just remove the code you added in your fix pr

pkg/importer/importer.go
```diff
@@ -533,6 +529,17 @@ func (h *Host) InitHandler(hl blobserver.FindHandlerByTyper) error {
 	h.targetStorage = rh.Storage
 	h.blobSource = rh.Storage
 
+	h.targetStorage, _ = h.target.(blobserver.Storage)
+	if h.targetStorage != nil {
+		var err error
+		h.targetClient, err = client.New(client.OptionUseStorageClient(h.targetStorage))
+		if err != nil {
+			log.Printf("Raindrop importer: error creating client: %v", err)
```
bradfitz: can't say "raindrop" in pkg/importer/*.go (not raindrop-specific)

pkg/importer/importer.go
bradfitz: Yeah, I haven't traced through it yet but seems like the `h.target` (StatReceiver) that gets configured isn't a full blobserver.Storage.

I'm in the middle of family traveling chaos now and won't have time to look again for maybe 12 hours.

In any case, we should really make it the case that all importers have easy access to a client (of some sort) to easily inject stuff into the right place, without each importer having to jump through hoops to do it themselves.

PR #1792, opened 2026-01-25
apps: initial webdav

MichaHoffmann:
Initial WebDAV support as an app. Mostly written by claude but I did test it using PhotoSync and cadaver and it worked just fine!

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -21,15 +21,19 @@ limitations under the License.
 package app // import "perkeep.org/pkg/app"
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 
 	"perkeep.org/internal/httputil"
 	"perkeep.org/pkg/auth"
+	"perkeep.org/pkg/blob"
 	"perkeep.org/pkg/client"
+	"perkeep.org/pkg/schema"
 )
 
 // Auth returns the auth mode for the app to access Perkeep, as defined by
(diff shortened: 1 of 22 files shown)


pkg/app/app.go
```diff
@@ -86,3 +100,73 @@ func PathPrefix(r *http.Request) string {
...
+
+// Signer provides server-side signing for apps that don't have direct access
+// to the secret key ring.
+type Signer struct {
+	cl     *client.Client
+	server string
+}
+
+// NewSigner returns a Signer that uses the given client and server URL
+// for server-side signing via the /sighelper/ endpoint.
```
bradfitz, line 121: always /sighelper/ or does it do discovery to find the appropriate endpoint, I assume?

pkg/test/testdata/server-config.json
```diff
@@ -8,10 +8,11 @@
          "handler": "root",
          "handlerArgs": {
              "ownerName": "test",
-             "blobRoot": "/bs-and-maybe-also-index/",
+             "blobRoot": "/bs-recv/",
```
bradfitz, line 11: what's with all these changes throughout? Are these unrelated?

Or can some of these parts be sent out separately in their own PRs or at least their own commits in one PR?

It's hard for me to understand the rationale when it's all lumped together in one commit.

MichaHoffmann: Most of it was to make the end to end tests with the test.World work iirc! Let me see if I can restructure it a bit better.

pkg/app/app.go
MichaHoffmann: Yes - the comment is misleading; it doesn't know about the specific endpoint and uses the Client.

PR #1796, opened 2026-09-06
pkg/blobserver/proxycache: close the blob fetched from origin in Fetch

amwolff:
The origin reader was never closed, leaking a connection or file handle on every miss.


pkg/blobserver/proxycache/proxycache.go
```diff
@@ -160,13 +160,14 @@ func (sto *Storage) Fetch(ctx context.Context, b blob.Ref) (rc io.ReadCloser, si
...
 		log.Printf("warning: proxycache cache fetch error for %v: %v", b, err)
 	}
-	rc, size, err = sto.origin.Fetch(ctx, b)
+	og, size, err := sto.origin.Fetch(ctx, b)
 	if err != nil {
-		return
+		return nil, 0, err
 	}
-	all, err := io.ReadAll(rc)
+	defer og.Close()
```
tgulacsi: Move this after the ReadAll, without defer.

tgulacsi (changes requested):

tgulacsi (approved):

amwolff:
@tgulacsi do you know who's maintaining the perkeep project these days? is it still @bradfitz and @mpl?

tgulacsi:
AFAIK @mpl has stepped down, @bradfitz and @radkat was active last winter holiday, but since that I haven't seen any activity...

[shy-question]Soft fork?[/shy-question]

amwolff:
the fork idea is indeed tempting, but I wonder if @bradfitz / @radkat would be open to appointing someone still active on the project as a new maintainer. I think this would be more productive, though it would require passing the torch to someone new.

